PR #271, opened 2026-04-01
Fix/ story media details fixes

### LINKED ISSUE #267: [Bug][Backend] No media retrieval in `GET /stories/{id}/` endpoint
### Description

Currently, the backend supports uploading media (e.g., images) when creating stories, but there is no functionality to retrieve or display these media files. As a result, the frontend cannot access or render images associated with a story.

The `GET /stories/{id}/` endpoint should be extended to include media information, and the backend should provide a way to access and display full-size images.

### Scope

- Extend story retrieval endpoint to include associated media  
- Provide accessible URLs for media files  
- Enable full image viewing support  
- Ensure proper serialization of media data

### Current Behavior

- Media can be uploaded when creating a story  
- Media is not returned in `GET /stories/{id}/`  
- No accessible endpoint or URL for retrieving images  
- Frontend cannot display story images  

### Expected Behavior

- `GET /stories/{id}/` returns associated media information  
- Each media item includes a valid, accessible URL  
- Images can be displayed in the frontend  
- Full-size image viewing is supported via direct URL or dedicated endpoint  

### Requirements

- Media files are stored and served via a publicly accessible URL
- Response includes all media linked to the story
- Support multiple media items per story
- Ensure correct media type (e.g., image) is included
- URLs should be usable directly in  tags

### Steps to Reproduce (Current Issue)

- Create a story with an image
- Call GET /stories/{id}/
- Observe that:
- No media information is included
- No way to access uploaded image

### Why This Matters

- Core functionality: Stories often rely on visual content
- Frontend integration: Required to render images in story detail page
- User experience: Incomplete stories without media
- Feature completeness: Upload without retrieval is insufficient

### Acceptance Criteria

- [ ] GET /stories/{id}/ includes a media field with all associated media
- [ ] Each media item includes a valid, accessible URL
- [ ] Images can be successfully rendered in the frontend using returned URLs
- [ ] Multiple media items per story are supported
- [ ] Media URLs return the correct file when accessed directly
- [ ] Backend handles media serialization consistently

### Meta (optional)

_No response_

AysuKeskin:
# 📝 Description
Fixes #267 

This PR improves the story detail and feed responses by exposing additional frontend-needed fields that were missing from the API.

The main motivation for this change is a bug noticed on the frontend: images attached to stories were not being displayed on the story detail page. The root cause was that the story detail endpoint did not include the related media items in its serialized response, so the frontend had no structured way to access and render story images.

To fix this, this PR introduces a dedicated `StoryDetailSerializer` that extends the base story serializer and includes a nested `media_items` field. Each media item is returned with its `id`, `url`, `media_type`, and `order`, allowing the frontend to render uploaded story images directly from the detail response.

The media item serializer also builds absolute URLs when request context is available, so frontend clients do not need to reconstruct file URLs manually or depend on backend media path knowledge.

In addition to fixing the image rendering issue on the story detail page, this PR also adds `user_has_liked` and `user_has_saved` fields to story serializers. These fields allow the frontend to understand the authenticated user’s interaction state for a story without making extra requests.

The detail endpoint is updated to use `StoryDetailSerializer`, and related query optimizations were added using `select_related('user')` and `prefetch_related('media_items')` to avoid unnecessary database queries.

The feed and search serializers were also updated to receive request context so that interaction-dependent fields such as `user_has_liked` and `user_has_saved` can be computed correctly.

Finally, this PR adds serializer and view tests covering media serialization, empty/non-empty media cases, ordering, absolute URL generation, and authenticated interaction states.

### Changes made
- Added `StoryMediaItemSerializer` for nested media serialization
- Added `StoryDetailSerializer` extending `StorySerializer`
- Added `media_items` to story detail responses
- Returned media items with:
  - `id`
  - `url`
  - `media_type`
  - `order`
- Built absolute media URLs when request context is available
- Updated `StoryDetailView` to use `StoryDetailSerializer`
- Updated `StoryDetailView` queryset to use:
  - `select_related('user')`
  - `prefetch_related('media_items')`
- Added `user_has_liked` to `StorySerializer`
- Added `user_has_saved` to `StorySerializer`
- Added `user_has_liked` to `StoryFeedSerializer`
- Added `user_has_saved` to `StoryFeedSerializer`
- Passed `request` context into serializers in feed/search/detail flows
- Updated tests for:
  - story detail media field presence
  - empty media list behavior
  - uploaded image inclusion
  - ordering of media items
  - absolute URL generation
  - liked/saved state for authenticated and unauthenticated users

### Why this fix is needed
The immediate reason for this PR is a frontend-visible bug: story images were not appearing on the story detail page because the backend response did not expose media items.

By including `media_items` directly in the detail response:
- the frontend can render story images without extra API calls
- the API becomes more aligned with the data needs of the detail page
- media rendering logic becomes simpler and more reliable
- file URL handling becomes safer and more consistent

The additional `user_has_liked` and `user_has_saved` fields also improve frontend behavior by allowing the UI to render interaction state directly from the API response, instead of inferring it or fetching it separately.

Overall, this PR makes the story detail response more complete and frontend-friendly.

# 📊 Type of Change
- [x] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Documentation
 
Example expected detail response behavior:
- story with no media → `"media_items": []`
- story with uploaded images → `"media_items": [{...}]`
- authenticated user who liked story → `"user_has_liked": true`
- authenticated user who saved story → `"user_has_saved": true`

# ⏱️ Estimated Review Time
- [ ] < 5 min
- [ ] 5-15 min
- [x] > 15 min

### DIFF
--- a/backend/apps/stories/serializers.py
+++ b/backend/apps/stories/serializers.py
@@ -10,12 +10,14 @@ class StorySerializer(serializers.ModelSerializer):
     user = serializers.PrimaryKeyRelatedField(read_only=True)
     user_has_liked = serializers.SerializerMethodField()
     user_has_saved = serializers.SerializerMethodField()
+    contributor_name = serializers.SerializerMethodField()
 
     class Meta:
         model = Story
         fields = [
             'id',
             'user',
+            'contributor_name',
             'title',
             'narrative',
             'location_lat',
@@ -38,6 +40,7 @@ class StorySerializer(serializers.ModelSerializer):
         read_only_fields = [
             'id',
             'user',
+            'contributor_name',
             'like_count',
             'save_count',
             'user_has_liked',
@@ -73,6 +76,11 @@ class StorySerializer(serializers.ModelSerializer):
         if request and hasattr(request, 'user') and request.user.is_authenticated:
             return obj.saved_by.filter(user=request.user).exists()
         return False
+    def get_contributor_name(self, obj):
+        """Return the author's username, or None if they have chosen to post anonymously."""
+        if obj.contributor_visible and obj.user:
+            return obj.user.username
+        return None
 
     def validate_status(self, value):
         # Only moderation endpoints may set a story to removed — reject it here
--- a/backend/apps/stories/tests/test_serializers.py
+++ b/backend/apps/stories/tests/test_serializers.py
@@ -121,8 +121,9 @@ class TestStorySerializerFields:
     def test_serializer_contains_expected_fields(self, story):
         serializer = StorySerializer(story)
         expected = {
-            'id', 'user', 'title', 'narrative', 'location_lat', 'location_lng',
-            'location_name', 'region', 'time_type', 'year', 'year_start', 'year_end',
+            'id', 'user', 'contributor_name', 'title', 'narrative',
+            'location_lat', 'location_lng', 'location_name', 'region',
+            'time_type', 'year', 'year_start', 'year_end',
             'status', 'contributor_visible', 'like_count', 'save_count',
             'user_has_liked', 'user_has_saved', 'submitted_at', 'updated_at',
         }
@@ -134,6 +135,24 @@ class TestStorySerializerFields:
         assert serializer.is_valid(), serializer.errors
         assert 'like_count' not in serializer.validated_data
         assert 'save_count' not in serializer.validated_data
+
+    def test_contributor_name_returns_username_when_visible(self):
+        user = make_user()
+        story = make_story(user=user, contributor_visible=True)
+        data = StorySerializer(story).data
+        assert data['contributor_name'] == user.username
+
+    def test_contributor_name_is_none_when_not_visible(self):
+        user = make_user()
+        story = make_story(user=user, contributor_visible=False)
+        data = StorySerializer(story).data
+        assert data['contributor_name'] is None
+
+    def test_contributor_name_is_none_when_user_is_anonymized(self):
+        # Story whose author account was deleted — user FK is null
+        story = make_story(user=None, contributor_visible=True)
+        data = StorySerializer(story).data
+        assert data['contributor_name'] is None
 
 
 @pytest.mark.django_db


KmalM (commented):
Hey Aysu, great work on this PR — the media items fix is solid and the test coverage is thorough. I have one performance suggestion before we merge.

**N+1 queries on `user_has_liked` / `user_has_saved` in the feed**

In the current implementation, `get_user_has_liked` and `get_user_has_saved` each fire an individual `.filter(...).exists()` DB query per story. For a page of 10 stories that's 20 extra queries on top of the main one, and it scales linearly with page size.

This can be eliminated by annotating the queryset with Django's `Exists` expression in `StoryFeedView` and `StorySearchView` before passing it to the serializer:

```python
# views.py — inside StoryFeedView.get() and StorySearchView.get()
from django.db.models import Exists, OuterRef
from apps.interactions.models import Like, SavedStory

if request.user.is_authenticated:
    qs = qs.annotate(
        _user_has_liked=Exists(Like.objects.filter(story=OuterRef('pk'), user=request.user)),
        _user_has_saved=Exists(SavedStory.objects.filter(story=OuterRef('pk'), user=request.user)),
    )
```

Then update the serializer methods to read the pre-computed value when available, falling back to the DB only for the single-object detail case:

```python
def get_user_has_liked(self, obj):
    if hasattr(obj, '_user_has_liked'):
        return obj._user_has_liked
    request = self.context.get('request')
    if request and hasattr(request, 'user') and request.user.is_authenticated:
        return obj.likes.filter(user=request.user).exists()
    return False
```

This folds both `Exists` checks into the main SQL query, so the feed stays at 1 query regardless of page size. The detail view (`StoryDetailView`) is unaffected — it fetches a single object so the fallback `.exists()` fires at most once, which is fine.

Everything else looks good to me.

KmalM (changes requested):

AysuKeskin:
Thanks for the suggestion. You're right, the N+1 is an issue worth fixing.
                                                                                                            
  Applied the Exists + OuterRef approach:
                                                                                                            
  - Added annotate_user_interactions() helper in views.py that annotates the queryset with _user_has_liked  
  and _user_has_saved subqueries                                                                            
  - StoryFeedView and StorySearchView now call it when the user is authenticated                            
  - Both serializer methods check hasattr(obj, '_user_has_liked/saved') first and fall back to .exists()    
  only for the detail view (single object, fires at most once — acceptable)                                 
  - Added tests covering the annotation path for both StorySerializer and StoryFeedSerializer               
                                                                                                            
  Feed stays at 1 query regardless of page size for authenticated users. Ready for re-review.

PR #291, opened 2026-04-02
fix/frontend story card refinement

### LINKED ISSUE #263: [Bug][Frontend] Story card does not limit visible character count
### Description

The story cards in the Story Feed display the full story content without any character limit or truncation. When a story contains a long text, it overflows the intended card layout, causing inconsistent card sizes and negatively affecting readability and UI consistency.

### Steps to Reproduce

- Navigate to the Story Feed page
- Create a story with a long text content if there is none
- Observe the story card in the feed

### Actual Behavior

- The full story text is displayed without truncation
- Cards become uneven in height
- Layout appears broken or visually inconsistent

### Expected Behavior

- Story content should be truncated after a certain number of characters or lines and it should not overflow from the card view.
- The full narrative should be visible when the story card is clicked and story detail page is opened.

### Why This Matters 

- UI Consistency: Uneven card sizes break the visual structure of the feed
- Readability: Long unbroken text reduces scannability
- User Experience: Users should quickly browse multiple stories without distraction
- Responsiveness: Layout issues become more noticeable on smaller screens

### Acceptance Criteria

- [ ] Story text in cards is truncated to a defined limit (e.g., character clamp)
- [ ] All story cards maintain consistent height and width regardless of content length, no overflows.
- [ ] Truncated text ends with an ellipsis (...) or similar indicator
- [ ] No layout breaking occurs with long text inputs
- [ ] Behavior is consistent across different screen sizes and devices

### Meta (optional)

Effort: S

sezindogan:
# 📝 Description
Fixes #263 

Refines the StoryCard component with several improvements and bug fixes. 

#### Story preview truncation fix
Discovered that a line clamp at 20 words was already in place, but it did not handle cases where a word exceeded the limit or caused awkward truncation. This resulted in a bug where the preview text could be cut off mid-word.
Fixed the issue by introducing a `truncateAtWord` utility. Instead of cutting strictly at the character limit, it walks back to the nearest word boundary before appending an ellipsis. This ensures that preview text always ends on a complete word and maintains readability.

  #### Contributor name
  The contributor's username is now displayed below the story title, sourced from story.contributor_name which is already returned by the feed endpoint.

  #### User has liked indicator
  A heart icon is now shown in the top-right corner of each card. It renders filled red when the authenticated user has liked the story (story.user_has_liked = true), and grey otherwise. This field is already available in the feed serializer response.

  #### Like count and comment count -> deferred
  Like count and comment count cannot be displayed at this time as neither field is included in the feed serializer response. This will be reconsidered after MVP once the backend exposes these values.

  #### Media indicator -> deferred, existing code removed
  The media badge has been removed from the card. The feed endpoint does not return any media information (media_items is only available on the story detail endpoint), so there is no data to drive the indicator. This will be reconsidered after MVP pending a backend change to expose a  has_media flag on the feed serializer.

# 📊 Type of Change
- [x]  Bug fix
- [x]  New feature
- [ ]  Refactoring
- [ ]  Documentation
 
# 📸 Screenshots / Recordings
Screenshot 2026-04-02 at 16 55 53
Screenshot 2026-04-02 at 16 55 53

# ⏱️ Estimated Review Time
- [x] < 5 min
- [ ] 5-15 min
- [ ] > 15 min

### DIFF
--- a/.env.prod.example
+++ b/.env.prod.example
@@ -1,0 +1,21 @@
+DJANGO_SETTINGS_MODULE=config.settings.production
+SECRET_KEY=change-me-to-a-50-char-random-string
+ALLOWED_HOSTS=yourdomain.com
+CORS_ALLOWED_ORIGINS=https://yourdomain.com
+DB_NAME=historystorymap
+DB_USER=dbuser
+DB_PASSWORD=change-me
+DB_ROOT_PASSWORD=change-me-root
+DB_HOST=127.0.0.1
+DB_PORT=3306
+MEDIA_ROOT=/app/media
+# Passed as a Docker build arg to nginx — the React bundle will call this URL.
+VITE_API_URL=https://yourdomain.com/api
+# Injected into the nginx config at container startup via envsubst.
+NGINX_HOST=yourdomain.com
+# Optional: set to your Sentry DSN to enable error tracking.
+SENTRY_DSN=
+EMAIL_HOST=smtp.example.com
+EMAIL_PORT=587
+EMAIL_HOST_USER=
+EMAIL_HOST_PASSWORD=
--- a/backend/Dockerfile.prod
+++ b/backend/Dockerfile.prod
@@ -1,0 +1,51 @@
+# ── Stage 1: Build ────────────────────────────────────────────────────────────
+# Install all build-time C dependencies and compile Python packages into a venv.
+# gcc and pkg-config are only needed here and are not carried into the final image.
+FROM python:3.12-slim AS builder
+
+RUN apt-get update && apt-get install -y \
+    default-libmysqlclient-dev \
+    gcc \
+    pkg-config \
+    libmagic1 \
+    && rm -rf /var/lib/apt/lists/*
+
+RUN python -m venv /venv
+ENV PATH="/venv/bin:$PATH"
+
+COPY requirements /requirements
+RUN pip install --no-cache-dir -r /requirements/production.txt
+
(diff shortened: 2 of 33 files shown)


sezindogan:
Thanks for the review!                                                                                                                        
                        
  On the truncation approach: The original triple-layer strategy was intentional, truncateAtWord(80) handled a long-word-no-space overflow edge case and added an explicit … to signal truncation, while line-clamp-3 served as a visual safety net. The concern with relying on line-clamp-3 alone was that when the 20-word backend preview fits within 3 lines (which it often does), there'd be no visible indicator that the story continues.                                                                                                                                    
                  
  After re-examining, though, the break-words class already handles the long-word overflow, so truncateAtWord wasn't actually needed for that. And truncateAtWord(80) doesn't reliably add … either — if the backend preview is short-worded (say, 60 chars), the 80-char cutoff never triggers and the text appears complete with no indication it's truncated.                                                                     
                  
  The cleaner fix is to always append … unconditionally since preview_text is always a truncated excerpt by definition. Dropped truncateAtWord  from the component and now render {preview}… directly, with line-clamp-3 + break-words handling visual overflow. This guarantees the truncation indicator regardless of preview length.                                                                                            
                  
  Other changes:
  - Added user_has_liked: false as an explicit default in makeStory() to make the contract clear
  - Removed the extra blank line left over from the CardFooter removal                                                                          
  - Updated the preview test to reflect the new behavior (no longer passing a preview that already ends with …)

PR #340, opened 2026-04-07
added lnt and lat for get stories/feed and get stories/map

### LINKED ISSUE #274: [Feature][Backend] Bounding-box filtering on search and filter endpoints
### Description

Stories already store location_lat and location_lng. Add optional bounding-box query parameters to /stories/feed/ and /stories/map/ so the frontend can filter stories that fall within a geographic rectangle (provided by a geocoder such as Nominatim).                               
   
No model changes or migrations are needed, this is purely additional filtering logic in the existing views.

### Acceptance Criteria

- [ ] /stories/feed/ accepts optional lat_min, lat_max, lng_min, lng_max query params                                                           
- [ ]  /stories/map/ accepts the same params                                                                                                       
- [ ]  Stories are filtered to those whose coordinates fall within the bbox when params are present
- [ ]  Params are independently optional (e.g. only lat_min can be supplied)                                                                       
- [ ]  Invalid values (non-numeric, lat_min > lat_max) return HTTP 400                                                                             
- [ ]  Bbox filtering composes correctly with existing year_from, year_to, and location filters (AND logic)                                        
- [ ]  Unit tests cover: bbox filtering, invalid param validation, combination with existing filters  

### Meta (optional)

_No response_

AHMET-CAGDAS-GIRIT:
# 📝 Description
Fixes # (issue number)

# 📊 Type of Change
- [ ]  Bug fix
- [X]  New feature
- [ ]  Refactoring
- [ ]  Documentation
 
- [X] Project builds successfully
- [X] My code follows the style guidelines of this project
- [X] I have commented my code, especially in hard-to-understand parts
- [X] I have performed a self-review of my code
- [X] I have added/updated tests
- [X] New and existing unit tests pass locally with my changes

# 📸 Screenshots / Recordings

# ⏱️ Estimated Review Time
- [ ] < 5 min
- [X] 5-15 min
- [ ] > 15 min

### DIFF
--- a/backend/apps/stories/serializers.py
+++ b/backend/apps/stories/serializers.py
@@ -330,6 +330,12 @@ class FeedQuerySerializer(serializers.Serializer):
     # Maximum of 500 km — uncapped radius would force a full table scan into memory.
     radius_km = serializers.FloatField(required=False, min_value=0.001, max_value=500.0)
 
+    # Bounding box filter — all four are independently optional
+    lat_min = serializers.FloatField(required=False, min_value=-90, max_value=90)
+    lat_max = serializers.FloatField(required=False, min_value=-90, max_value=90)
+    lng_min = serializers.FloatField(required=False, min_value=-180, max_value=180)
+    lng_max = serializers.FloatField(required=False, min_value=-180, max_value=180)
+
     def validate(self, data):
         year_from = data.get('year_from')
         year_to = data.get('year_to')
@@ -345,6 +351,18 @@ class FeedQuerySerializer(serializers.Serializer):
             raise serializers.ValidationError(
                 {f: 'Required when performing radius filtering.' for f in missing}
             )
+
+        if data.get('lat_min') is not None and data.get('lat_max') is not None:
+            if data['lat_min'] > data['lat_max']:
+                raise serializers.ValidationError(
+                    {'lat_max': 'lat_max must be greater than or equal to lat_min.'}
+                )
+
+        if data.get('lng_min') is not None and data.get('lng_max') is not None:
+            if data['lng_min'] > data['lng_max']:
+                raise serializers.ValidationError(
+                    {'lng_max': 'lng_max must be greater than or equal to lng_min.'}
+                )
 
         return data
 
--- a/backend/apps/stories/tests/test_views.py
+++ b/backend/apps/stories/tests/test_views.py
@@ -223,6 +223,44 @@ class TestStoryFeedView:
         response = client.get(FEED_URL + '?tag=folklore&location=Istanbul')
         assert response.data['count'] == 1
         assert response.data['results'][0]['title'] == 'Match'
+    
+    def test_bbox_non_numeric_lat_min_returns_400(self, client):
+        response = client.get(FEED_URL + '?lat_min=abc')
+        assert response.status_code == 400
+        
+    def test_bbox_non_numeric_lng_max_returns_400(self, client):
+        response = client.get(FEED_URL + '?lng_max=xyz')
+        assert response.status_code == 400
+    
+    def test_bbox_lat_min_greater_than_lat_max_returns_400(self, client):
+        response = client.get(FEED_URL + '?lat_min=42&lat_max=40')
+        assert response.status_code == 400
+    
+    def test_bbox_lng_min_greater_than_lng_max_returns_400(self, client):
+        response = client.get(FEED_URL + '?lng_min=30&lng_max=20')
+        assert response.status_code == 400
+        
+    def test_bbox_only_lat_min_does_not_crash(self, client):
+        response = client.get(FEED_URL + '?lat_min=40')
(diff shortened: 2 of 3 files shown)


AysuKeskin (changes requested):
Hey, the logic itself is fine but a few things caught my eye.       

---                                                            
                                                                                                                                   
  backend/__init__.py and backend/test_settings.py — why are these here? backend/ isn't a Python package, that __init__.py does nothing there. And test_settings.py is dead code — pytest reads config/settings/test.py, it never touches this file. Just delete both.                                                                                                                              
                  
--- 
                                                                                                            
  Wrong URL in the map test — there's a bbox test inside TestStoryMapView but look at this:                                          
   
  response = client.get(FEED_URL, params)  # should be MAP_URL                                                                       
                                                                                                                                     
  This test isn't testing the map view at all, it's hitting the feed. It passes but it's giving you false confidence — the map bbox path has zero actual test coverage.                  

---                                                                              
                                                                                                                                     
  No negative tests — does lat_min > lat_max actually return a 400? The validation is there in the serializer but nothing is         
  asserting it. Add something like:
                                                                                                                                     
  def test_invalid_lat_range_returns_400(self, client):
      response = client.get(MAP_URL + '?lat_min=42.0&lat_max=40.0')
      assert response.status_code == status.HTTP_400_BAD_REQUEST                                                                     
   
  Same for lng, and cover both feed and map.    

---                                                                                     
                  
  Rest of it looks good.

AHMET-CAGDAS-GIRIT:
I modified the tests and removed the redundant files. Should be working now

AysuKeskin (changes requested):
Overall the implementation looks solid. The bbox filter logic itself is correct .
                                   
  A few things still need to be addressed before this is ready to merge:                                       
  
  ---                                                                                                    
  Bug: Wrong URLs in tests
                          
  There are three tests where the URL being called doesn't match the class it lives in. This means those
  specific scenarios are not actually being tested for the endpoint they claim to cover.                 
  
  1. TestStoryFeedView.test_feed_bbox_filtering (line 208) — this test is inside the Feed test class, but
   it calls MAP_URL. It should call FEED_URL to actually verify that the feed endpoint's bbox filtering
  works.                                                                                                 
  2. TestStoryMapView.test_bbox_and_year_filter_combined and
  TestStoryMapView.test_bbox_and_location_combined (lines 511 and 520) — these are inside the Map test   
  class, but both call FEED_URL. They should call MAP_URL.
                                                                                                         
  These look like copy-paste mistakes. The tests pass because both endpoints share the same filter logic,
   but they're testing the wrong endpoint.
                                                                                                         
  ---             
  Missing file: backend/apps/__init__.py
                                        
  Can you please add back the __init__.py file under backend/apps/? It needs to be there for Django's
  migration system to correctly discover app modules — without it, manage.py migrate and makemigrations  
  can fail to resolve the apps package when run from the backend/ directory. It's an empty file, but it
  marks apps/ as a Python package which Django relies on.                                                
                  
  ---
  Once those three test lines are fixed and the __init__.py is restored, this is good to go.

AHMET-CAGDAS-GIRIT:
should be done now

AysuKeskin (dismissed):
Thanks for the effort on addressing the requested changes.

However, this PR has grown significantly in scope. It currently includes a large number of unrelated commits (including merges from other PRs and features not directly related to the original purpose of this PR), which makes it very difficult to review and reason about safely. 

When addressing review feedback, we should aim to keep the PR focused only on the relevant changes. Pulling in many unrelated commits introduces noise and increases the risk of unintended side effects.

At this point, I would recommend:
- either rebasing/cleaning the branch to include only the relevant changes for this PR,  
- or closing this PR and opening a clean one with a minimal, focused diff. (You can use cherry-pick to take related commits from this branch.)

This will make the review process much more reliable and maintainable.

AHMET-CAGDAS-GIRIT:
Ok. I pulled the latest main and pasted the implementations I made to that latest version. conflict should be far fewer now

AysuKeskin:
**Important Note:** The content and commit message of this commit db1422b25a35f3cb6ac1d8dd12152c0c3b87261d are not mine. Since this branch was significantly behind `main`, I had to rebase it onto the latest `main`. During the rebase, Git recreated the commit object, so GitHub now shows it as `authored by AHMET-CAGDAS-GIRIT, committed by AysuKeskin`. The original author and owner of the change is AHMET-CAGDAS-GIRIT; my only involvement was updating the branch on top of the latest `main`.

AysuKeskin:
@KmalM can you please review the latest version of PR and close it if everything looks good? Since I was involved in creating the latest version of it, It would be reasonable to review it by someone else.

KmalM (approved):
Cosmetic issues noted but non-blocking. Logic is correct, tests are thorough, diff is clean. Good to merge.

KmalM:
All set!

PR #344, opened 2026-04-07
fix: add automatic token refresh interceptor

### LINKED ISSUE #343: [Bug][Frontend] Handle Automatic Token Refresh
### Description

Currently, users are logged out after access token expiration (~15 minutes) because the frontend does not attempt to refresh the token.

This issue aims to fix this behavior by integrating the refresh token mechanism into the frontend, allowing seamless session continuation without forcing users to log in again.

---

## 🎯 Objective

Ensure uninterrupted user sessions by automatically refreshing expired access tokens.

---

## ⚙️ What needs to be implemented

1. **Token Storage**
   - Store access & refresh tokens securely  
   - Ensure persistence across sessions (e.g., AsyncStorage / localStorage)

2. **API Interceptor**
   - Add request/response interceptor (e.g., Axios interceptor)  
   - Detect `401 Unauthorized` responses caused by expired access token  

3. **Refresh Flow**
   - On token expiration:
     - Send request to `/auth/refresh`
     - Replace access token with the new one
     - Retry the original request  

4. **Failure Handling**
   - If refresh token is invalid/expired:
     - Log out user  
     - Redirect to login screen  

5. **Concurrency Handling (Important)**
   - Prevent multiple simultaneous refresh requests  
   - Queue pending requests while refreshing token  

6. **Logout Integration**
   - Clear both access & refresh tokens on logout  

---

## 📊 Type of Change
- [x] Bug fix  
- [ ] New feature  
- [ ] Refactoring  
- [ ] Documentation  

---

### Acceptance Criteria

- [ ] Access token expiration does NOT log out user immediately  
- [ ] Expired access token triggers refresh flow automatically  
- [ ] Original request is retried successfully after refresh  
- [ ] Invalid refresh token logs out user  
- [ ] Tokens are stored and updated correctly  
- [ ] No duplicate refresh requests occur  
- [ ] Works on both mobile and web (if applicable)  

---

### Meta (optional)

_No response_

omarreis01:
# Description
Fixes #343

Users were being logged out immediately when their access token expired (~15 min in prod) because the frontend's API interceptor cleared tokens and redirected to `/login` on any 401 response. The refresh token was stored but never used.

This PR replaces the simple 401 handler with a refresh-aware interceptor that:
1. Attempts to renew the access token via `POST /auth/refresh/` before logging out
2. Retries the original failed request with the new token
3. Deduplicates concurrent refresh calls (multiple 401s → single refresh request)
4. Guards against infinite retry loops with a `_retry` flag
5. Skips refresh for auth endpoints (`/auth/login/`, `/auth/refresh/`, `/auth/logout/`)
6. Falls back to clear & redirect if refresh fails or no refresh token exists

**Note:** This requires the backend to expose `POST /auth/refresh/` (simplejwt `TokenRefreshView`). Until that endpoint is available, the interceptor gracefully falls back to the existing logout behavior.

# Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Refactoring
- [ ] Documentation

# Screenshots / Recordings
N/A — logic-only change (no UI changes)

# Estimated Review Time
- [ ] < 5 min
- [x] 5-15 min
- [ ] > 15 min

### DIFF
--- a/frontend/src/services/api.js
+++ b/frontend/src/services/api.js
@@ -54,7 +54,7 @@ api.interceptors.response.use(
 
     // Skip refresh for auth endpoints — just clear and redirect
     const requestUrl = originalRequest.url || "";
-    if (AUTH_ENDPOINTS.some((ep) => requestUrl.endsWith(ep))) {
+    if (AUTH_ENDPOINTS.some((ep) => requestUrl.includes(ep))) {
       clear();
       navigationRef.navigate?.("/login");
       return Promise.reject(error);
@@ -70,6 +70,7 @@ api.interceptors.response.use(
     originalRequest._retry = true;
 
     // Attempt to refresh the token (dedup concurrent calls)
+    let refreshedData;
     try {
       if (!refreshPromise) {
         refreshPromise = axios
@@ -85,15 +86,17 @@ api.interceptors.response.use(
       }
 
       const { data } = await refreshPromise;
-
-      // Retry the original request with the new token
-      originalRequest.headers.Authorization = `Bearer ${data.access}`;
-      return api(originalRequest);
+      refreshedData = data;
     } catch {
       clear();
       navigationRef.navigate?.("/login");
       return Promise.reject(error);
     }
+
+    // Retry the original request with the new token — let non-auth errors
+    // propagate to the caller instead of incorrectly logging out.
+    originalRequest.headers.Authorization = `Bearer ${refreshedData.access}`;
+    return api(originalRequest);
   }
 );
 


sezindogan:
## PR Review                                                                                                                                  
                  
  - Deduplication (refreshPromise) is correctly implemented — concurrent 401s produce exactly one refresh request, which the test for "only makes one 
  refresh request for concurrent 401s" verifies.
  - _retry guard prevents infinite loops. The flag is set before the try block, so even if the retry is intercepted again (e.g., server keeps         
  returning 401), the second invocation hits the guard and clears out.                                                                                
  - Token rotation is handled — setRefreshToken(res.data.refresh) stores the new refresh token that SimpleJWT rotates on each use.
  - Auth endpoint bypass prevents refresh attempts on login/logout/refresh itself.                                                                    
  - Test coverage is comprehensive — happy path, refresh failure, dedup, infinite retry, and auth endpoint skipping are all covered.                  
                                                                                                                                                      
###  Issues to address                                                                                                                                   
                                                                                                                                                      
  1. catch block is too broad (minor bug)                                                                                                             
   
  api.js:92                                                                                                                                           
                  
  } catch {                                                                                                                                           
    clear();      
    navigationRef.navigate?.("/login");
    return Promise.reject(error);
  }

  The catch covers both "refresh call failed" and "retry request threw a non-401 error" (e.g., network error on the retried request). A network error after a successful refresh will incorrectly log the user out and discard the newly-acquired tokens. Split the two cases:
                                                                                                                                                      
  let refreshedData;
  try {
    // ...refresh logic...
    refreshedData = data;                                                                                                                             
  } catch {
    clear();                                                                                                                                          
    navigationRef.navigate?.("/login");
    return Promise.reject(error);
  }

  originalRequest.headers.Authorization = `Bearer ${refreshedData.access}`;                                                                           
  return api(originalRequest); // let caller handle non-auth errors normally
                                                                                                                                                      
  2. endsWith URL matching (api.js:57)                                                                                                                
   
  if (AUTH_ENDPOINTS.some((ep) => requestUrl.endsWith(ep))) {                                                                                         
                                                                                                                                                      
  Works correctly for relative paths (which is how axios stores them internally), but would silently break if a query param is ever appended to an auth endpoint — endsWith("/auth/login/") fails for /auth/login/?next=.... includes is safer here, though query params on these endpoints are currently impossible.

omarreis01:
Thanks for the review @sezindogan! Both issues are addressed in 0855920:

1. **Catch block split** — The try/catch now only wraps the refresh logic. The retried request (`api(originalRequest)`) is outside the catch, so non-auth errors (e.g. network errors after a successful refresh) propagate normally to the caller instead of incorrectly clearing tokens and redirecting to login.

2. **`endsWith` → `includes`** — Auth endpoint matching now uses `includes` so it stays correct even if query params are appended (e.g. `/auth/login/?next=...`).

All 9 existing tests still pass.

PR #357, opened 2026-04-11
feat: add automated test report generation and wiki publishing

### LINKED ISSUE #356: [Task][DevOps] Add automated test report generation and wiki publishing
### Description

The CI pipeline runs backend, frontend, and mobile test suites but does not produce any persistent, human-readable test report. This task adds an automated workflow that runs all three test suites after every push to `main`, generates a structured Markdown report (with pass/fail summaries and per-test detail tables), and publishes it to the wiki page "Generated Unit Test Reports".

- Backend tests run with `pytest-json-report` → JSON
- Frontend tests run with `vitest --reporter=json` → JSON
- Mobile tests run with `jest --json` → JSON
- A Python script converts all three JSON outputs into a single formatted Markdown wiki page

### Acceptance Criteria

- [ ] A `Generate Test Reports` workflow exists in `.github/workflows/test-reports.yml`
- [ ] Workflow runs on every push to `main` and on manual dispatch
- [ ] Backend, frontend, and mobile tests each produce a JSON report
- [ ] A Python script combines the three reports into a single Markdown page
- [ ] The page is pushed automatically to the wiki as `Generated Unit Test Reports`
- [ ] The report shows pass/fail summary per suite and an expandable full test list
- [ ] Report is generated even when some tests fail (`continue-on-error: true`)

### Meta (optional)

Effort: M
Priority: P2

KmalM:
# 📝 Description
Fixes #356

Adds a `Generate Test Reports` workflow that runs all three test suites (backend, frontend, mobile) on every push to `main`, generates structured JSON output from each, and publishes a formatted Markdown report to the wiki page **Generated Unit Test Reports**.

The report shows a pass/fail summary per suite and an expandable full test list with per-test status and duration. Each test suite runs with `continue-on-error: true` so the report is always generated and pushed even when tests fail.

Implementation:
- `.github/workflows/test-reports.yml` — orchestrates the full pipeline; uses `pytest-json-report` (backend), `vitest --reporter=json` (frontend), `jest --json` (mobile), then pushes to the wiki via `GITHUB_TOKEN`
- `.github/scripts/generate_test_report.py` — reads the three JSON outputs and builds the wiki Markdown page with summary lines and `<details>` test tables

The wiki page stub has already been created at: https://github.com/bounswe/bounswe2026group4/wiki/Generated-Unit-Test-Reports

# 📊 Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Refactoring
- [x] Documentation

# 📸 Screenshots / Recordings

# ⏱️ Estimated Review Time
- [ ] < 5 min
- [x] 5-15 min
- [ ] > 15 min

### DIFF
--- a/.github/scripts/generate_test_report.py
+++ b/.github/scripts/generate_test_report.py
@@ -1,7 +1,14 @@
 #!/usr/bin/env python3
 """
 Read pytest-json-report (backend), vitest --reporter=json (frontend),
-and jest --json (mobile) output files and generate a single Markdown wiki page.
+and jest --json (mobile) output files and generate separate Markdown wiki
+pages for each suite plus an index page.
+
+Output files (written to GITHUB_WORKSPACE):
+  Backend-Generated-Unit-Test-Reports.md
+  Frontend-Generated-Unit-Test-Reports.md
+  Mobile-Generated-Unit-Test-Reports.md
+  Generated-Unit-Test-Reports.md  (index linking to the three above)
 """
 
 import json
@@ -13,6 +20,8 @@ SERVER_URL = os.environ.get("GITHUB_SERVER_URL", "https://github.com")
 REPOSITORY = os.environ.get("GITHUB_REPOSITORY", "")
 RUN_ID = os.environ.get("GITHUB_RUN_ID", "")
 RUN_URL = f"{SERVER_URL}/{REPOSITORY}/actions/runs/{RUN_ID}" if RUN_ID else ""
+
+WIKI_BASE = f"{SERVER_URL}/{REPOSITORY}/wiki"
 
 
 def _icon(outcome):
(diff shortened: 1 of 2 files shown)


AysuKeskin (approved):
This looks good overall. Automating test report generation and publishing the results to the wiki is a valuable addition for the project, especially since it improves visibility of test outcomes and makes the testing process much easier to track over time.

I also think this is particularly useful for the MVP review and documentation phase, because having structured and regularly published test reports strengthens the project’s transparency and makes it easier to demonstrate testing coverage and project maturity.

The decision to keep report generation running even when some suites fail is also a strong choice, since it ensures that the team can still inspect results instead of losing visibility when a workflow breaks. Overall, this seems like a practical and well-timed contribution to the repository.

AysuKeskin:
From the current CI logs, I noticed something we could not catch locally and may have assumed was working correctly: 
the workflow fails at the `cp` step because `Backend-Generated-Unit-Test-Reports.html` is not found at the expected path. This suggests that the HTML report generation step is not actually producing the file successfully, or it is writing it with a different filename or to a different location. Since the copy step depends on that file, it fails with `cannot stat`. It would be good to verify that the HTML generation step always creates the report at the exact expected path, with matching filename casing, before the copy step runs.

Can you please fix that?

PR #363, opened 2026-04-14
feat: implement auto login after registration

### LINKED ISSUE #362: [Bug][Frontend] Auto-Authenticate User After Registration
### Description

Currently, users are redirected to the login page immediately after completing registration, which adds unnecessary friction to the onboarding flow.
Implement automatic authentication immediately after successful registration so that users are logged in without needing to sign in manually. After successful registration and automatic login:

- If the user was redirected to register from another page or action, they should be returned to their original intended destination/state.
- If the user navigated to register from the login page, they should still be redirected back to their original intended destination rather than the login page.
- If no previous destination exists, redirect the user to the default post-registration flow (e.g. profile completion page or home page).

This should ensure a seamless onboarding experience and preserve user context throughout the authentication flow.

### Acceptance Criteria

- [ ] After successful registration, the user is automatically authenticated.
- [ ] No additional manual login step is required after registration.
- [ ] The authenticated session is established correctly and persists across page refresh/navigation.
- [ ] If automatic authentication fails, the user is shown a clear error message and guided to log in manually.
- [ ] The flow works correctly on desktop and mobile devices.
- [ ] Existing registration validations and error handling remain unaffected.
- [ ] Unit tests are modified or added accordingly.

### Meta (optional)

_No response_

sezindogan:
# 📝 Description
Fixes #362 

Improved the registration flow by adding automatic authentication after successful sign-up and preserving the user’s intended destination across login/register redirects.

### Registration Flow

* Added `useLocation` import in `frontend/src/pages/RegisterPage.jsx`
* Added `useAuth` hook import in `RegisterPage.jsx`
* After successful `register()`, automatically calls `login(email, password)` using `AuthContext`
* On successful auto-login:

  * Shows success toast: **"Account created! Welcome!"**
  * Redirects user to:

    * `location.state?.from?.pathname` if available, or
    * `/` as fallback
* On auto-login failure:

  * Shows fallback toast: **"Account created! Please sign in to continue."**
  * Redirects to `/login`
  * Preserves original intended destination in navigation state

### Login Page

* Updated the **Sign up** link in `frontend/src/pages/LoginPage.jsx`
* Passes:

  * `state={{ from: location.state?.from }}`
* Ensures original intended destination (e.g. `/submit-story`) is preserved when moving:

  * protected route → login → register → auto-login → intended page

### Tests

Updated `frontend/src/pages/__tests__/RegisterPage.test.jsx`:

* Replaced `BrowserRouter` with `MemoryRouter` to support `location.state`
* Added `useAuth` mock with `mockLogin`
* Added tests for:

  * auto-login is called after successful registration
  * redirect to home when no previous destination exists
  * redirect to original destination when `from` exists
  * success welcome toast is shown
  * auto-login failure fallback redirects to `/login` with preserved `from` state
  * fallback guidance toast is shown
* Updated existing success-path tests to include `mockLogin`
* Removed obsolete test:

  * redirect to `/login` after successful registration

## Result

Users are now automatically signed in after registration and correctly redirected to their intended destination, creating a smoother onboarding experience.

# 📊 Type of Change
- [x]  Bug fix
- [ ]  New feature
- [ ]  Refactoring
- [ ]  Documentation
 
# 📸 Screenshots / Recordings

# ⏱️ Estimated Review Time
- [x] < 5 min
- [ ] 5-15 min
- [ ] > 15 min

### DIFF
--- a/frontend/src/components/AppLayout/AppLayout.jsx
+++ b/frontend/src/components/AppLayout/AppLayout.jsx
@@ -93,10 +93,10 @@ function AppLayout() {
                 </>
               ) : (
                 <>
-                  <Button variant="ghost" size="sm" onClick={() => navigate("/login")}>
+                  <Button variant="ghost" size="sm" onClick={() => navigate("/login", { state: { from: location } })}>
                     Login
                   </Button>
-                  <Button size="sm" onClick={() => navigate("/register")}>
+                  <Button size="sm" onClick={() => navigate("/register", { state: { from: location } })}>
                     Register
                   </Button>
                 </>
@@ -153,14 +153,14 @@ function AppLayout() {
                         variant="ghost"
                         size="sm"
                         className="justify-start"
-                        onClick={() => { setSheetOpen(false); navigate("/login"); }}
+                        onClick={() => { setSheetOpen(false); navigate("/login", { state: { from: location } }); }}
                       >
                         Login
                       </Button>
                       <Button
                         size="sm"
                         className="justify-start"
-                        onClick={() => { setSheetOpen(false); navigate("/register"); }}
+                        onClick={() => { setSheetOpen(false); navigate("/register", { state: { from: location } }); }}
                       >
                         Register
                       </Button>
--- a/frontend/src/pages/LoginPage.jsx
+++ b/frontend/src/pages/LoginPage.jsx
@@ -183,6 +183,7 @@ function LoginPage() {
             Don&apos;t have an account?{" "}
             <Link
               to="/register"
+              state={{ from: location.state?.from }}
               className="font-medium text-primary underline-offset-4 hover:underline"
             >
               Sign up
--- a/frontend/src/pages/RegisterPage.jsx
+++ b/frontend/src/pages/RegisterPage.jsx
@@ -1,5 +1,5 @@
 import { useState } from "react";
-import { useNavigate, Link } from "react-router-dom";
+import { useNavigate, useLocation, Link } from "react-router-dom";
 import { Mail, Lock, Eye, EyeOff, Loader2, MapPin, User, Check, X } from "lucide-react";
 
 import {
@@ -14,6 +14,7 @@ import {
   CardFooter,
 } from "@/components/ui";
 import { register } from "@/services/authService";
+import { useAuth } from "@/hooks/useAuth";
 import { useToast } from "@/hooks/useToast";
 
 const INITIAL_FIELD_ERRORS = {
@@ -44,6 +45,8 @@ function validatePasswordStrength(password) {
 
 function RegisterPage() {
   const navigate = useNavigate();
+  const location = useLocation();
+  const { login } = useAuth();
   const { toast } = useToast();
 
   const [username, setUsername] = useState("");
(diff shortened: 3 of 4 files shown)


omarreis01 (approved):
## PR Review: feat: implement auto login after registration

### Summary
Cleanly implements auto-login after successful registration with a graceful fallback, and plumbs the `from` location through the Login ↔ Register toggle so a protected-route redirect survives the register flow. Pure frontend change, well-tested. Lint clean and all 28 RegisterPage tests pass locally.

---

### Frontend

**`RegisterPage.jsx`** (`frontend/src/pages/RegisterPage.jsx:144-171`)
- Nested try/catch is the right shape: inner error only catches auto-login failure, outer still handles register failure and always runs `setIsLoading(false)` in `finally`.
- Destructuring `from` into `{ pathname, search, hash }` before passing to `navigate` is the correct pattern — it preserves query strings (e.g. map filters `?q=castle&year_from=1900`) and hash anchors, which the PR explicitly tests for.
- `{ replace: true }` on both success and fallback navigation is correct — the user shouldn't be able to Back-button into the register page with prefilled fields.
- `toast.info` (vs `toast.success`) for the fallback case is a good UX signal that something didn't fully go to plan.

**`LoginPage.jsx`** (`frontend/src/pages/LoginPage.jsx:186`)
- Forwarding `location.state?.from` on the Sign-up `<Link>` is minimal and correct. `location` was already destructured at line 21, so no new import needed.

**`AppLayout.jsx`** (`frontend/src/components/AppLayout/AppLayout.jsx:93-163`)
- Passing `{ state: { from: location } }` on all four Login/Register entry points (desktop + mobile) means the redirect chain works from any navbar entry, not just from a protected route.

**Test coverage** (`frontend/src/pages/__tests__/RegisterPage.test.jsx`)
- Switch to `MemoryRouter` is necessary to seed `location.state` — correct call.
- New tests cover: auto-login invocation, redirect-to-home when no `from`, redirect-to-`from` when present, query-param preservation, success toast, fallback navigation on auto-login failure, and fallback toast. Solid coverage.
- The obsolete "navigates to /login on successful registration" test was correctly removed rather than kept as a stale assertion.

---

### Backend

No backend changes. The PR relies on the existing `loginService` (via `AuthContext.login`, see `frontend/src/context/AuthContext.jsx:29-34`), which is already in production use from `LoginPage`.

---

No new endpoints or contract changes. The auto-login path reuses the same `login(email, password)` call that `LoginPage` already uses — behavior is identical from the backend's perspective (one extra `POST /auth/login/` immediately after `POST /auth/register/`).

---

### Non-blocking notes

1. **Two network round-trips for registration.** The current flow does `register` → `login`, sending the password twice. Not a security concern over HTTPS, but a cleaner pattern is for the register endpoint to return tokens directly (single round-trip, and immune to the edge case below). Worth a follow-up ticket, not a blocker.
2. **Edge case: normalization drift.** If `register` and `login` normalize email differently (e.g. one trims/lowercases and the other doesn't), auto-login could silently fail and the user lands on `/login`. The fallback handles this gracefully — just flagging it as something to watch if you see the fallback toast fire in production analytics.
3. **Minor style:** the inline conditional in the `navigate(...)` call is a bit dense. A one-line helper (`const target = from ? { pathname: from.pathname, search: from.search, hash: from.hash } : "/"`) would make it easier to scan, but entirely optional.

---

### Verdict
**APPROVE ✅**

Nice, tight change that meaningfully improves onboarding UX. No blockers.

PR #364, opened 2026-04-14
Fix/search filter params

### LINKED ISSUE #359: [Bug][Backend] Fix inability to combine search and filter parameters
### Description

There is currently a backend inconsistency between the story feed endpoint and the story search endpoint.

At the moment, the search endpoint does not accept the same filtering parameters supported by the feed endpoint. Because of this, users cannot apply search and filtering together in a single request. In practice, this means that once a search query is used, filters such as the ones available on the feed side cannot be applied on top of those search results, which breaks the expected discovery flow.

This is a bug because search and filtering are meant to work together as part of the story discovery experience. A user should be able to search for a story and then further narrow the results using the same filtering options already supported by the feed endpoint.

The backend should be updated so that the search endpoint supports the relevant filter parameters already handled by the feed endpoint, and applies them together with the search query in a consistent way.

This change should also ensure that the request/response behavior remains aligned across discovery-related endpoints, so the frontend can use search and filtering in combination without needing separate workarounds.

**Expected behavior:**

* Users can send a search query and filter parameters in the same request
* The search endpoint applies both the text search and the provided filters
* Filtering behavior is consistent with the feed endpoint
* Search + filter combinations return correctly narrowed results

**Current behavior:**

* The search endpoint only handles the search query
* Feed-style filter parameters are not accepted or not applied
* Search and filtering cannot be used together

### Acceptance Criteria

* [ ] The search endpoint accepts the same relevant filtering parameters as the feed endpoint
* [ ] Search and filtering can be used together in a single request
* [ ] Combined search + filter requests return correct results
* [ ] Backend behavior is consistent between feed and search endpoints
* [ ] Existing search functionality continues to work correctly
* [ ] Tests are added or updated for search + filter combinations

### Meta (optional)

Estimated Duration: 2 hours
Deadline: 18-04-2026

AysuKeskin:
# 📝 Description
Fixes #359

This PR fixes the backend bug where the story search endpoint could not accept the same filter parameters as the feed endpoint, which prevented search and filtering from being used together.

With this change, the search endpoint now supports the required `q` parameter together with optional feed-style filters such as `sort_by`, `year_from`, `year_to`, and `location`.

The PR includes:
- extending `SearchQuerySerializer` to validate these additional filter parameters
- adding validation for invalid year ranges
- updating the search service to apply the filters
- updating the view to pass validated parameters correctly
- adding and updating serializer, service, and view tests for valid, invalid, and combined filter cases

# 📊 Type of Change
- [x]  Bug fix
- [ ]  New feature
- [ ]  Refactoring
- [ ]  Documentation
 
# 📸 Screenshots / Recordings

# ⏱️ Estimated Review Time
- [ ] < 5 min
- [x] 5-15 min
- [ ] > 15 min

### DIFF
--- a/backend/apps/stories/serializers.py
+++ b/backend/apps/stories/serializers.py
@@ -298,7 +298,31 @@ class StoryMapSerializer(serializers.ModelSerializer):
 
 
 class SearchQuerySerializer(serializers.Serializer):
-    """Validates the q query parameter for the story search endpoint."""
+    """
+    Validates query parameters for the story search endpoint.
+
+    q is required. The filter params (sort_by, year_from, year_to, location) are
+    optional and mirror those accepted by FeedQuerySerializer so search results
+    can be narrowed with the same filters as the feed.
+    """
+
+    SORT_RECENT = 'recent'
+    SORT_POPULAR = 'popular'
+    SORT_CHOICES = [SORT_RECENT, SORT_POPULAR]
 
     # strip_whitespace=True (default) means a whitespace-only value becomes '' and fails min_length
     q = serializers.CharField(required=True, min_length=1)
+    sort_by = serializers.ChoiceField(choices=SORT_CHOICES, default=SORT_RECENT, required=False)
+    year_from = serializers.IntegerField(required=False)
+    year_to = serializers.IntegerField(required=False)
+    # Substring match against location_name — partial values like "galata" are valid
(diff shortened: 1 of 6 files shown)


KmalM (approved):
The implementation is logically sound. Year filtering correctly handles both single-year stories (year field) and year-range stories (year_start/year_end), consistent with how get_story_feed does it. Much needed addition.

PR #396, opened 2026-04-17
Fix/ hide author_username in comments when user has private username

### LINKED ISSUE #393: [Bug][backend]  Hide private usernames in story comments on backend responses
## Description

There is a privacy gap in the current comment flow. On mobile, story contributor names can be anonymized using public profile data, but comment authors cannot be anonymized reliably for other users because the comment API response only includes `author_username` and does not provide enough information to determine whether that username should be hidden.

Right now, when a user makes their username private, other users may still see that username in story comments because the backend continues returning it in comment responses.

This should be fixed on the backend so privacy is enforced at the source of truth. The preferred behavior is:

- if a comment author has a private username, the backend should not expose their username in comment responses
- instead, the response should return `author_username: null` so clients can consistently render `Anonymous`

This avoids requiring extra client-side profile lookups and ensures consistent privacy behavior across mobile, web, and any future clients.

## Testing

- [ ] Add/update serializer tests for comments authored by users with `is_username_public = false`.
- [ ] Add/update view/API tests to confirm private users return `author_username: null`.
- [ ] Add/update tests to confirm public users still return their username.
- [ ] Confirm anonymized/deleted-author comment behavior still works as expected.

## Notes

- This issue should be solved in the backend, not with client-side workarounds.
- Returning `author_username: null` is preferred over exposing extra identifiers and requiring clients to resolve privacy manually.

### Acceptance Criteria

- [ ] If a user's `is_username_public` setting is `false`, comment API responses do not expose that user's username.
- [ ] For private users, comment responses return `author_username: null` instead of the real username.
- [ ] If a comment is already anonymized or the author is missing, behavior remains correct and unchanged.
- [ ] If a user's `is_username_public` setting is `true`, their username continues to be returned normally in comment responses.
- [ ] The change applies consistently to all relevant comment endpoints.
- [ ] Existing comment creation, listing, and deletion behavior remains unaffected.
- [ ] Backend tests are added or updated accordingly.

### Meta (optional)

_No response_

AysuKeskin:
# 📝 Description
Fixes #393 

This PR updates comment serialization so that `author_username` is hidden when the comment author has a private username setting (`is_username_public=False`).

It also adds serializer and view tests to verify:
- private usernames are returned as `null`
- public usernames are still shown normally
- anonymized comment behavior remains unchanged

# 📊 Type of Change
- [x]  Bug fix
- [ ]  New feature
- [ ]  Refactoring
- [ ]  Documentation
 
# ⏱️ Estimated Review Time
- [x] < 5 min
- [ ] 5-15 min
- [ ] > 15 min

### DIFF
--- a/backend/apps/interactions/serializers.py
+++ b/backend/apps/interactions/serializers.py
@@ -22,6 +22,8 @@ class CommentResponseSerializer(serializers.ModelSerializer):
     def get_author_username(self, obj):
         if obj.is_anonymized or obj.author is None:
             return None
+        if not obj.author.is_username_public:
+            return None
         return obj.author.username
 
 
--- a/backend/apps/interactions/tests/test_serializers.py
+++ b/backend/apps/interactions/tests/test_serializers.py
@@ -60,6 +60,25 @@ class TestCommentResponseSerializer:
         for field in ['id', 'story_id', 'author_username', 'text', 'is_anonymized', 'created_at']:
             assert field in data
 
+    def test_response_serializer_hides_username_when_private(self, story):
+        private_user = User.objects.create_user(
+            email='private@example.com', username='privateuser', password='Password1',
+            is_username_public=False,
+        )
+        comment = Comment.objects.create(story=story, author=private_user, text='Incognito')
+        data = CommentResponseSerializer(comment).data
+        assert data['author_username'] is None
+        assert data['is_anonymized'] is False
+
+    def test_response_serializer_shows_username_when_public(self, story):
+        public_user = User.objects.create_user(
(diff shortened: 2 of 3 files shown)


KmalM (approved):
In `test_views.py` both new view tests do:
```python
assert response.data['results'][0]['author_username'] is ...
```
  

  If the story fixture or Django's default ordering ever returns an existing comment before the one just created, the assertion will     
  target the wrong comment. A safer approach:
```python
comments = response.data['results']
  target = next(c for c in comments if c['text'] == 'Hidden identity')
  assert target['author_username'] is None
```

However it is not a merge blocker, so approved! Nicely done.

PR #398, opened 2026-04-17
Feat/ add first_name, last_name, and name visibility to user profile

### LINKED ISSUE #397: [Bug][Backend] Add visibility-controlled first name and last name fields to user profile
### Description

A frontend request revealed a gap in the current user profile implementation: at the moment, the backend only exposes `username` as profile identity information, and there are no dedicated **first name / last name** fields that users can provide and control the visibility of.

This creates a mismatch with the expected profile behavior, since users may want to display their real name separately from their username and decide whether that information should be publicly visible. The current backend structure does not support that use case yet.

To address this, the user profile model and related API responses should be extended to support:

* a **first name** field,
* a **last name** field,
* and corresponding **visibility controls** so users can decide whether these fields are public or private.

This request came from the frontend side, where support for configurable name visibility is needed as part of profile display and profile editing flows.

The implementation should ensure that:

* these fields can be stored and updated,
* their visibility preferences are respected in public profile responses,
* and profile edit/view endpoints remain consistent with the rest of the profile visibility logic.

### Acceptance Criteria

* [ ] A first name field is added to the relevant user/profile model structure
* [ ] A last name field is added to the relevant user/profile model structure
* [ ] Visibility control is supported for both fields
* [ ] Profile update endpoints allow users to edit these fields and their visibility settings
* [ ] Public profile responses respect the visibility settings for first name and last name
* [ ] Existing profile behavior continues to work correctly
* [ ] Tests are added or updated for first name / last name visibility behavior

### Meta (optional)

Estimated Duration: 1 hour

AysuKeskin:
# 📝 Description
Fixes #397 

This PR adds `first_name`, `last_name`, and `is_name_public` support to user profiles. This feature was requested from the frontend side and helps app to track user informations better.

Changes included in this PR:
- Added `first_name` and `last_name` fields to `UserProfile`
- Added `is_name_public` visibility flag to control whether name fields are shown publicly
- Updated profile serializers to accept and return the new fields
- Updated public profile serialization to hide/show `first_name` and `last_name` based on `is_name_public`
- Ensured the profile owner can still see their own name fields even when `is_name_public` is set to `False`
- Added migration for the new profile fields
- Added model, serializer, and view tests covering the new functionality

# 📊 Type of Change
- [ ]  Bug fix
- [x]  New feature
- [ ]  Refactoring
- [ ]  Documentation
 
# 📸 Screenshots / Recordings

# ⏱️ Estimated Review Time
- [x] < 5 min
- [ ] 5-15 min
- [ ] > 15 min


KmalM (changes requested):
## Review

The feature logic, serializer structure, and test coverage all look good — the implementation follows existing patterns correctly. However, there is one **deployment-blocking issue** that must be fixed before this can merge.

---

### ❌ Migration modified in place (blocker)

`0002_user_is_username_public_user_total_points_and_more.py` was edited to inject the new fields into an existing `CreateModel` operation, rather than generating a new migration file.

**Why this breaks deployments:**
Django tracks which migrations have already run by recording their filename in the `django_migrations` table. Every team member's local DB, CI, and any shared environment that already applied `0002` will **never** pick up the new `first_name`, `last_name`, and `is_name_public` columns — Django sees `0002` as done and skips it entirely. The result is an `OperationalError: column userprofile.first_name does not exist` crash at runtime.

**Fix:** Revert `0002` to its original content and generate a new migration instead:

```bash
docker compose run --rm --entrypoint="" web sh -c \
  "python manage.py makemigrations users --name userprofile_name_fields"
```

This produces `0003_userprofile_name_fields.py`, which applies cleanly on all existing environments.

---

### Minor notes (non-blocking)

- `test_patch_name_fields_updates_in_db` and `test_patch_is_name_public_updates_in_db` don't assert `response.status_code == 200` — the tests would pass silently even if the PATCH returned a 400.
- `from apps.users.models import UserProfile` is imported inside individual test methods; it should be at the top of the file with the other imports.

Everything else looks solid. Once the migration is corrected, this is ready to merge.

AysuKeskin:
-  Migration modified in place (blocker)
                                                                                                             
  You're right that the migration was wrong, but to clarify: 0002 was not manually edited to inject fields   
  into the CreateModel — what happened was 0002 was deleted and regenerated from scratch, which caused Django
   to bake the new fields directly into the CreateModel operation. The end result was the same deployment    
  problem you described. Fixed by restoring 0002 to its original content and running makemigrations users 
  --name userprofile_name_fields to produce a proper 0003_userprofile_name_fields.py with three AddField
  operations depending on 0002.

-  Missing status_code assertions

  Valid — added assert response.status_code == status.HTTP_200_OK to both                                    
  test_patch_name_fields_updates_in_db and test_patch_is_name_public_updates_in_db.
                                                                                                             
 -  Inline UserProfile imports inside test methods

  This follows the existing convention in the file — all other test methods in TestCurrentUserView and       
  TestPublicUserProfileView use the same inline import style for UserProfile. We kept it consistent with that
   pattern rather than diverging within a single class.

KmalM (approved):
The migration blocker is fixed — `0003_userprofile_name_fields.py` is a proper new migration chained off `0002`, and both PATCH tests now assert the response status code. All previous concerns are addressed. ✅

PR #402, opened 2026-04-17
feat: extend media upload endpoint to support audio and video

### LINKED ISSUE #225: [Feature][Backend] Extend media upload endpoint to support audio and video files
### Description

The platform currently supports image uploads for stories via the \`MediaItem\` model, which already has \`media_type\` choices for IMAGE, AUDIO, and VIDEO. This issue extends the upload endpoint to accept audio (mp3, wav, ogg) and video (mp4, webm) file types. Uploaded files should be stored using the existing \`MediaItem\` infrastructure and return a playable URL for use by frontend and mobile media players.

**Scope:**

* extend the story media upload endpoint to accept audio and video files
* accept audio formats: mp3, wav, ogg
* accept video formats: mp4, webm
* store accepted files using the existing \`MediaItem\` model with the appropriate \`media_type\`
* return the media item id, media_type, and playable file URL in the response
* update the upload serializer to handle multi-type media input
* add service-layer logic for audio/video file handling
* return 400 for unsupported file types
* return 403 when a non-owner attempts to upload
* register the endpoint under the stories route
* add serializer, service, and endpoint tests

**Suggested branch name:**
\`feat/audio-video-upload\`

**Depends on:**
\`feat/image-upload-endpoint\` (issue: #172)

### Acceptance Criteria

* [ ] The upload endpoint accepts mp3, wav, and ogg audio files
* [ ] The upload endpoint accepts mp4 and webm video files
* [ ] Uploaded files are stored via \`MediaItem\` with the correct \`media_type\`
* [ ] Response includes the media item id, media_type, and accessible file URL
* [ ] Unsupported file types return a 400 error
* [ ] Non-owners receive a 403 response
* [ ] URL is wired under the stories route
* [ ] Serializer, service, and endpoint tests are added and passing

### Meta (optional)

Deadline: 15-04-2026
Estimated Time: 3 hours

KmalM:
# 📝 Description
Fixes #225

Extends the story media upload infrastructure to support audio (mp3, wav, ogg) and video (mp4, webm) files via a new `POST /stories/<id>/media/` endpoint. The existing `POST /stories/<id>/images/` endpoint is unchanged.

**Changes:**
- `serializers.py` — new `MediaFileUploadSerializer`: accepts any `FileField`, uses `python-magic` to detect MIME from raw bytes (spoofing-resistant), maps MIME → `MediaType.AUDIO/VIDEO`, enforces per-type size limits (50 MB audio / 200 MB video). Also adds `media_type` to `MediaItemResponseSerializer`.
- `services.py` — new `upload_story_media(story, file, media_type)` that mirrors `upload_story_image` but accepts the media type as a parameter.
- `views.py` + `urls.py` — new `StoryMediaUploadView` at `POST /stories/<id>/media/`, identical auth/ownership flow as the image endpoint (`IsOwnerOrAdmin`, removed stories → 404).
- Tests — serializer unit tests (all 5 formats, size limits, image rejection, `media_type` in `validated_data`), service unit tests (audio + video), view integration tests (success for audio + video, 401/403/404, unsupported type → 400, metadata stored correctly).

# 📊 Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Documentation

# 📸 Screenshots / Recordings

# ⏱️ Estimated Review Time
- [ ] < 5 min
- [x] 5-15 min
- [ ] > 15 min

### DIFF
--- a/backend/apps/media/serializers.py
+++ b/backend/apps/media/serializers.py
@@ -1,10 +1,22 @@
 import magic
 from rest_framework import serializers
 
-from apps.media.models import MediaItem
+from apps.media.models import MediaItem, MediaType
 
 ALLOWED_MIME_TYPES = {'image/jpeg', 'image/png'}
 MAX_FILE_SIZE = 2 * 1024 * 1024  # 2 MB
+
+ALLOWED_AUDIO_MIME_TYPES = {'audio/mpeg', 'audio/wav', 'audio/x-wav', 'audio/ogg'}
+ALLOWED_VIDEO_MIME_TYPES = {'video/mp4', 'video/webm'}
+ALLOWED_MEDIA_MIME_TYPES = ALLOWED_AUDIO_MIME_TYPES | ALLOWED_VIDEO_MIME_TYPES
+
+MIME_TO_MEDIA_TYPE = {
+    **{mime: MediaType.AUDIO for mime in ALLOWED_AUDIO_MIME_TYPES},
+    **{mime: MediaType.VIDEO for mime in ALLOWED_VIDEO_MIME_TYPES},
+}
+
+MAX_AUDIO_FILE_SIZE = 50 * 1024 * 1024   # 50 MB
+MAX_VIDEO_FILE_SIZE = 200 * 1024 * 1024  # 200 MB
 
 
 class ImageUploadSerializer(serializers.Serializer):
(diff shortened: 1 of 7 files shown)


AysuKeskin (commented):
Nice work overall — the serializer/service/view separation is clean, and the MIME sniffing approach is a solid security choice.

One thing I’d suggest revisiting is the use of `self._detected_mime` on the serializer instance. Storing request-specific state on the serializer like this can be a bit fragile, since serializers are generally expected to remain stateless in usage patterns like this. Even if it works in practice here, attaching the detected MIME type directly to the uploaded file object (or passing it through a safer intermediate structure) would make the flow more explicit and reduce the risk of unexpected state-related issues.

Something along these lines may be safer:

```python
def validate_file(self, value):
    ...
    detected_mime = magic.from_buffer(value.read(1024), mime=True)
    value.seek(0)
    value._detected_mime = detected_mime
    return value

def validate(self, attrs):
    attrs['media_type'] = MIME_TO_MEDIA_TYPE[attrs['file']._detected_mime]
    return attrs
````

Not blocking, but I think this would be a cleaner and safer pattern.

AysuKeskin (changes requested):
I think the current limits (50 MB for audio and 200 MB for video) are a bit too high for a community storytelling platform.

A more reasonable default could be:
- **Audio: 10 MB** — this should still cover around 5 minutes at 128 kbps for compressed formats like MP3 or OGG. WAV is uncompressed, so longer recordings would not fit, but that seems acceptable since users can convert those files if needed.
- **Video: 50 MB** — this should be enough for a short 2–3 minute clip at around 720p H.264, which feels more aligned with the expected use case here.

This would help keep storage usage more predictable and also reduce the risk of the platform being used more like a general media hosting service than a storytelling app.

Of course, if there is a concrete product need for larger uploads, we can revisit these values later.

KmalM (commented):
Fixed in b77bf79.

Size limits lowered: audio 50 MB → 10 MB, video 200 MB → 50 MB.

Also applied the `value._detected_mime` refactor from the first review — stores the detected MIME on the file object instead of `self`, avoiding per-request state on the serializer instance.

All 65 tests pass.

AysuKeskin (approved):
Looks good to me. The detected MIME is no longer stored on the serializer instance, and the upload limits were adjusted to the requested 10 MB for audio and 50 MB for video. The endpoint/service/test flow also looks consistent. 
Approving.

PR #408, opened 2026-04-18
fix: frontend search filter UX improvements

### LINKED ISSUE #406: [Bug][Frontend] Improve Search and Filter UX via a Default Value, Bounds and Persistence
The current implementation of search and filtering suffers from two primary UX friction points: inefficient input defaults and state volatility during view transitions.

- Year Filter Inefficiency: The year picker defaults to 0, forcing excessive manual navigation to reach contemporary dates. Additionally, the lack of an upper bound allows for logically invalid selections (e.g., year 3000).

- State Loss: Currently, search queries and filter parameters are reset when a user toggles between the Map and Feed views. This disrupts the user flow, requiring redundant inputs when switching perspectives on the same data.

This issue aims to centralize filter logic and enforce realistic constraints to improve overall usability and data integrity.

### Proposed Changes

Year Picker Optimization:
- Initialize the lower default year to 1980 and upper default year to 2026.
- Enforce a hard floor at 1000 and ceiling at 2030 (end of the current decade) to prevent illogical inputs.

Global State Persistence:

- Lift the search and filter state to a shared parent component or global store.
- Ensure that any active search string or year filter remains applied when navigating between MapView and FeedView.

### Acceptance Criteria

- [ ] Upon initialization, the "Year" field defaults to 1980.
- [ ] The year input is capped at 2030. The increment button is disabled once this value is reached.
- [ ] Manual entries exceeding the range (e.g., 2050) must auto-correct to the nearest bound (2030) or trigger a validation message.
- [ ] State of the search and filter parameters are persisted between views. 
- [ ] Existing unit tests for the filter component are updated.
- [ ] Integration tests for the state persistence between the two views are added.

sezindogan:
# 📝 Description
Fixes #406 
Improves the year filter UX and persists search/filter state when switching between the Feed and Map pages.                                          
                                                                                                                                                       
 Year filter changes: The From and To year fields now start empty (placeholder visible). Clicking the spinner starts From at 1980 and To at 2026. Values are capped at a maximum of 2030 and a minimum of 1000; typing above the max auto-corrects to 2030.                                            
                                                                                                                                                       
  Filter persistence: Navigating between Feed and Map via the navbar now carries the active query params (search term, year range, location) across    
  pages.
### Files Changed 
- src/components/SearchFilter/FilterPanel.jsx             
- src/components/SearchFilter/__tests__/FilterPanel.test.jsx
- src/components/AppLayout/AppLayout.jsx                                                                                                             
- src/components/AppLayout/__tests__/AppLayout.test.jsx

# 📊 Type of Change
- [x]  Bug fix
- [ ]  New feature
- [ ]  Refactoring
- [ ]  Documentation
 
# 📸 Screenshots / Recordings

# ⏱️ Estimated Review Time
- [x] < 5 min
- [ ] 5-15 min
- [ ] > 15 min

### DIFF
--- a/frontend/src/components/AppLayout/AppLayout.jsx
+++ b/frontend/src/components/AppLayout/AppLayout.jsx
@@ -14,8 +14,8 @@ import {
 import { useAuth } from "@/hooks/useAuth";
 
 const publicLinks = [
-  { to: "/map", label: "Map" },
-  { to: "/", label: "Feed" },
+  { to: "/map", label: "Map", preserveSearch: true },
+  { to: "/", label: "Feed", preserveSearch: true },
   { to: "/submit-story", label: "Submit Story", icon: Plus },
 ];
 
@@ -49,7 +49,12 @@ function AppLayout() {
     navigate("/");
   };
 
-  const links = publicLinks;
+  const isFilterPage = location.pathname === "/" || location.pathname.startsWith("/map");
+  const links = publicLinks.map((link) =>
+    link.preserveSearch && isFilterPage
+      ? { ...link, to: `${link.to}${location.search}` }
+      : link
+  );
 
   return (
     <div className="min-h-screen">
--- a/frontend/src/components/AppLayout/AppLayout.jsx
+++ b/frontend/src/components/AppLayout/AppLayout.jsx
@@ -19,8 +19,9 @@ const publicLinks = [
   { to: "/submit-story", label: "Submit Story", icon: Plus },
 ];
 
-function NavLink({ to, label, icon: Icon, pathname, onClick }) {
-  const isActive = to === "/" ? pathname === "/" : pathname.startsWith(to);
+function NavLink({ to, basePath, label, icon: Icon, pathname, onClick }) {
+  const base = basePath ?? to;
+  const isActive = base === "/" ? pathname === "/" : pathname === base || pathname.startsWith(base + "/");
 
   return (
     <Link
@@ -49,10 +50,10 @@ function AppLayout() {
     navigate("/");
   };
 
-  const isFilterPage = location.pathname === "/" || location.pathname.startsWith("/map");
+  const isFilterPage = location.pathname === "/" || location.pathname === "/map";
   const links = publicLinks.map((link) =>
-    link.preserveSearch && isFilterPage
-      ? { ...link, to: `${link.to}${location.search}` }
+    link.preserveSearch && isFilterPage && location.search
+      ? { ...link, basePath: link.to, to: `${link.to}${location.search}` }
       : link
   );
 
@@ -70,8 +71,9 @@ function AppLayout() {
           <nav className="hidden md:flex items-center space-x-6">
             {links.map((link) => (
               <NavLink
-                key={link.to}
+                key={link.label}
                 to={link.to}
+                basePath={link.basePath}
                 label={link.label}
                 icon={link.icon}
                 pathname={location.pathname}
@@ -123,8 +125,9 @@ function AppLayout() {
                 <nav className="flex flex-col space-y-4 mt-4">
                   {links.map((link) => (
                     <NavLink
-                      key={link.to}
+                      key={link.label}
                       to={link.to}
+                      basePath={link.basePath}
                       label={link.label}
                       icon={link.icon}
                       pathname={location.pathname}
--- a/frontend/src/components/AppLayout/__tests__/AppLayout.test.jsx
+++ b/frontend/src/components/AppLayout/__tests__/AppLayout.test.jsx
@@ -145,6 +145,26 @@ describe("AppLayout", () => {
     expect(mapLinks[0]).toHaveAttribute("href", "/map");
   });
 
+  it("Map link is still highlighted when filters are active on /map", () => {
+    renderWithRouter("/map?q=galata&year_from=1980");
+
+    const mapLink = screen.getByText("Map");
+    const feedLink = screen.getByText("Feed");
+
+    expect(mapLink.className).toContain("text-primary");
+    expect(feedLink.className).toContain("text-muted-foreground");
+  });
+
+  it("Feed link is still highlighted when filters are active on /", () => {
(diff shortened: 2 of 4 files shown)


omarreis01 (commented):
## PR Review: fix: frontend search filter UX improvements

### Summary
Frontend-only PR with two independent changes: (1) year-filter UX polish in `FilterPanel` (empty-on-open, spinner defaults, clamp to 2030, min 1000), and (2) search/filter persistence across Feed ↔ Map via navbar links in `AppLayout`. Code is small and focused, and test coverage for new behaviors is solid. However, there is **one regression in the active-link highlighting** that should be fixed before merge.

---

### Frontend

**FilterPanel.jsx**
- Constants `YEAR_MIN/YEAR_MAX/YEAR_SPINNER_FROM/YEAR_SPINNER_TO` pulled to the top — clean.
- `onKeyDown` ArrowUp/ArrowDown on empty fields: clear and well-tested.
- The spinner-click heuristic `if (localYearFrom === \"\" && Number(val) === YEAR_MIN)` is clever but brittle: if a user pastes `1000` (or types `1000` from an empty field via IME/autofill) it will be silently rewritten to `1980`. Low-probability, worth noting.
- `YEAR_SPINNER_TO = 2026` is a hardcoded \"current year.\" It will go stale next year. Consider `new Date().getFullYear()` — zero runtime cost, no more yearly maintenance.
- Validation `(isNaN(from) || from < YEAR_MIN)` — since the input is `type=\"number\"`, `isNaN` is defensive but harmless.

**AppLayout.jsx**
- Declarative `preserveSearch` flag on link config is nice.
- 🚫 **Blocker — active-link highlighting regresses when filters are active.**
  `NavLink` computes `isActive` as:
  ```js
  const isActive = to === \"/\" ? pathname === \"/\" : pathname.startsWith(to);
  ```
  After this PR, `to` may be `/map?q=galata&year_from=1980` while `pathname` stays `/map`. Then `\"/map\".startsWith(\"/map?q=galata&year_from=1980\")` is `false`, so the Map link is no longer highlighted on `/map` when any filter is set. Same for Feed on `/` with filters (the `to === \"/\"` shortcut also fails because `to` becomes `/?q=...`).
  Two straightforward fixes:
  1. Pass the bare path (without search) as a separate prop to `NavLink` and use it for `isActive`, or
  2. Keep the `to` prop bare in the `links` array and pass the search-augmented value only to `<Link to={...}>`, so the active computation stays independent of query string.
- Using `link.to` as `key` (line 68) now changes whenever the URL search changes, causing `NavLink` to remount on every filter change. Minor, but if you refactor as above, use `link.label` or a stable id for the key.
- `location.pathname.startsWith(\"/map\")` would also match a hypothetical `/mapped`, `/map-settings`, etc. No such routes today — flagging only for future-proofing. `=== \"/map\"` or `startsWith(\"/map/\")` is safer.

**Tests**
- New tests are clear, well-named, and cover the happy paths and the main edge cases (empty defaults, arrow-key spinner, clamp, below-min error, cross-page param carry, non-filter pages).
- Missing: a test pinning the active-link behavior when a filter is present in the URL (would have caught the regression above).

---

### Backend

No backend changes. No new API endpoints called. Nothing to verify on the contract side.

---

N/A — this PR only changes URL query strings that are already consumed by existing endpoints (`q`, `year_from`, `year_to`, `location`). No contract shifts.

---

### Verdict
**REQUEST CHANGES**

Blockers (must fix before merge):
- Active-state highlighting for Map/Feed links is broken when filters are present in the URL — `NavLink.isActive` should compare against `pathname` vs. the bare path, not the search-augmented `to`. Add a regression test for this.

Non-blocking notes:
- Replace hardcoded `YEAR_SPINNER_TO = 2026` with `new Date().getFullYear()`.
- Spinner-click heuristic may false-positive on paste/autofill of `1000` — consider guarding with an `onWheel`/`onClick`/`onPointerDown` flag instead of the value heuristic, or accept the edge case and add a brief comment.
- `startsWith(\"/map\")` → `=== \"/map\"` (or `startsWith(\"/map/\")`) for future-proofing.
- Consider a stable `key` on the `NavLink` list (e.g. `link.label`) once `to` may contain search params.

Nice work overall — the intent is clear and the tests are a pleasure to read. Just the one highlighting bug to sort out. 👍

omarreis01 (changes requested):
Blocking issue: active-link highlighting regresses for Map/Feed when filters are present in the URL. Details in the review comment above.

sezindogan:
Thanks for the thorough review! Here's what was addressed:                                                                                           
                                                                                                                                                       
 ## Fixed:                                                                                                                                               
                                                                                                                                                       
  - Active-link highlighting (blocker) — NavLink now receives a separate basePath prop (the bare path, e.g. /map) used solely for the isActive check,  
  while to carries the search-augmented URL for navigation. This means highlighting is correct regardless of what's in the query string.
  - YEAR_SPINNER_TO = 2026 — replaced with new Date().getFullYear() so it stays current automatically.                                                 
  - Stable key on NavLink list — switched from link.to (which changes with search params) to link.label.                                               
  - startsWith("/map") scope — tightened to === "/map" in the isFilterPage check to avoid false matches on hypothetical /map-settings etc.             
  - Regression test added — two new tests pin the active-link behavior when filters are present in the URL (the exact scenario that was broken).       
                                                                                                                                                       
  Not changed:                                                                                                                                         
                                                                                                                                                       
  - Spinner-click heuristic (YEAR_MIN → YEAR_SPINNER_FROM) — the false-positive on paste/autofill of 1000 is acknowledged as a real (if low-probability) edge case. An onPointerDown flag approach would be more precise, but it adds meaningful complexity for a very unlikely scenario. The existing comment already documents the intent, so I've left this for a dedicated follow-up if it surfaces in practice.

omarreis01 (approved):
Thanks for the thorough fixes! The active-link highlighting regression is resolved via the `basePath` prop split, YEAR_SPINNER_TO now uses `new Date().getFullYear()`, keys are stable, and the `/map` match is tightened. New regression tests pin the behavior. LGTM — approved.

PR #410, opened 2026-04-18
refactor: make map endpoint RFC 7946-compliant

### LINKED ISSUE #379: [Refactor][Backend] Add GeoJSON support for the story map endpoint
### Description

As part of the planned standards integration work, the backend map data response should be extended to support the **GeoJSON** standard. GeoJSON is the standard geographic JSON format planned for the project’s map-related data exchange, and it fits our story map structure naturally because each story already has a geographic point and associated location metadata. 

Right now, the story map endpoint returns data in the project’s current custom JSON structure. While this works for our current implementation, it makes the frontend depend on project-specific transformation logic for map rendering. Supporting GeoJSON on the backend would make the response more standard-compliant, improve interoperability, and make it easier for the frontend to consume map data in a cleaner and more reusable way. This also aligns with the standard adoption plan documented for the project.  

This task is specifically about the backend side of that change. The goal is to provide a valid GeoJSON representation of published stories on the map, where each story is returned as a geographic feature with the correct geometry and basic story/location metadata inside `properties`. The implementation can either be done by adding a dedicated GeoJSON route or by extending the existing endpoint with an alternative GeoJSON response format, depending on what fits the current API structure better. 

The response should follow the GeoJSON `FeatureCollection` structure and represent each story as a `Feature` with:

* a `Point` geometry,
* coordinates in the correct GeoJSON order,
* and relevant properties such as title, location name, and time-related fields. 

This issue is only for the backend response shape. Frontend consumption of the GeoJSON output should be handled separately.

### Acceptance Criteria

* [ ] The story map endpoint exposes a valid GeoJSON response, either through a dedicated route or a supported alternative response format
* [ ] The GeoJSON response follows the `FeatureCollection` structure
* [ ] Each published story is returned as a GeoJSON `Feature`
* [ ] Each feature contains `Point` geometry with coordinates in `[longitude, latitude]` order
* [ ] Each feature includes the relevant story metadata under `properties`
* [ ] Draft or removed stories are excluded from the GeoJSON output
* [ ] The API documentation/schema is updated if needed to reflect the new response format
* [ ] Tests are added or updated for the GeoJSON response behavior

### Meta (optional)

Estimated Duration: 2–3 hours

AysuKeskin:
# 📝 Description
Fixes #379 

This PR refactors the map endpoint to return an RFC 7946-compliant GeoJSON response instead of the previous custom paginated structure.

### What changed
- Replaced the old `StoryMapSerializer` with `StoryMapGeoJSONSerializer`
- Updated `GET /stories/map/` to return a GeoJSON `FeatureCollection`
- Ensured each story is serialized as a GeoJSON `Feature`
- Moved coordinates into `geometry.coordinates` using RFC 7946 ordering: `[longitude, latitude]`
- Moved map-related metadata into `properties`
- Removed pagination from the map endpoint so map clients can receive the full pin set in one response
- Updated serializer and view tests to match the new GeoJSON response contract
- Added tests verifying:
  - `FeatureCollection` response shape
  - `Feature` structure
  - `[lng, lat]` coordinate ordering
  - numeric coordinate serialization
  - filtering behavior with the new response format
  - no unintended field exposure in map properties

### Important note for frontend
Frontend map implementations should also be updated to follow this new response format as soon as possible. The endpoint no longer returns the old paginated structure with flat story fields, so any frontend code still expecting fields like `results`, `location_lat`, or `location_lng` at the top level will misread the backend response and likely fail in the map flow. To stay compatible with the backend, the frontend should now read map data from the GeoJSON structure (`features`, `geometry`, and `properties`) and align its parsing/rendering logic with this standard.

# 📊 Type of Change
- [ ] Bug fix
- [ ] New feature
- [x] Refactoring
- [ ] Documentation
 
# ⏱️ Estimated Review Time
- [ ] < 5 min
- [x] 5-15 min
- [ ] > 15 min

### DIFF
--- a/backend/apps/stories/serializers.py
+++ b/backend/apps/stories/serializers.py
@@ -272,13 +272,16 @@ class FeedQuerySerializer(serializers.Serializer):
         return data
 
 
-class StoryMapGeoJSONSerializer(serializers.ModelSerializer):
+class StoryMapGeoJSONSerializer(serializers.BaseSerializer):
     """
     Serializes a Story as a GeoJSON Feature (RFC 7946).
 
     Coordinates follow the RFC 7946 §3.1.1 mandate: [longitude, latitude].
     Decimal fields are cast to float so the JSON encoder emits numeric values,
     not quoted decimal strings.
+
+    BaseSerializer is used intentionally — we own the full output shape via
+    to_representation, so ModelSerializer's field introspection is pure overhead.
     """
 
     def to_representation(self, instance):
@@ -299,10 +302,6 @@ class StoryMapGeoJSONSerializer(serializers.ModelSerializer):
                 "year_end": instance.year_end,
             },
         }
-
-    class Meta:
-        model = Story
-        fields = []
 
 
 class SearchQuerySerializer(serializers.Serializer):
--- a/backend/apps/stories/serializers.py
+++ b/backend/apps/stories/serializers.py
@@ -272,29 +272,37 @@ class FeedQuerySerializer(serializers.Serializer):
         return data
 
 
-class StoryMapSerializer(serializers.ModelSerializer):
+class StoryMapGeoJSONSerializer(serializers.ModelSerializer):
     """
-    Read-only serializer returning only the fields needed to render a map pin.
+    Serializes a Story as a GeoJSON Feature (RFC 7946).
 
-    Intentionally minimal — the map view only needs coordinates, place name,
-    title, and time info to position a pin and show a preview tooltip.
-    Full story content is loaded separately when a pin is clicked.
+    Coordinates follow the RFC 7946 §3.1.1 mandate: [longitude, latitude].
+    Decimal fields are cast to float so the JSON encoder emits numeric values,
+    not quoted decimal strings.
     """
+
+    def to_representation(self, instance):
+        return {
+            "type": "Feature",
+            "id": instance.id,
+            "geometry": {
+                "type": "Point",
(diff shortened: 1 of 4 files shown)


KmalM (changes requested):
Overall clean, correct refactor with good test coverage.

One real issue — wrong base class:
                                                                                                                                           StoryMapGeoJSONSerializer still inherits from ModelSerializer despite overriding to_representation entirely and declaring Meta.fields =   []. DRF's ModelSerializer runs field introspection and field-generation machinery on every instantiation, which is all wasted work    
  here. The correct base class is serializers.Serializer (or serializers.BaseSerializer if you want even less overhead):                 

  
```python
# before
class StoryMapGeoJSONSerializer(serializers.ModelSerializer):
      ...
      class Meta:
          model = Story
          fields = []

  # after
  class StoryMapGeoJSONSerializer(serializers.BaseSerializer):
      def to_representation(self, instance):                                                                                                       return { ... }

```
BaseSerializer has no Meta, needs no model, and does zero field introspection — it's the canonical choice when you own the full output  shape.

AysuKeskin:
You’re right. `ModelSerializer` still goes through its usual field introspection even with `fields = []`, so `BaseSerializer` is the more appropriate base here.

It wasn’t a major issue since this was just a quick fix and the serializer was already working as intended, but your point is valid and it’s cleaner to correct it before merge. I’ve updated it in the same branch.

KmalM (approved):

PR #420, opened 2026-04-19
feat: implement follow/unfollow endpoints

### LINKED ISSUE #399: [Task][Backend] Implement follow/unfollow: model, endpoints, and event emission
## Description
Implement the core follow/unfollow write-side for the social graph feature.

Tasks:
- Create the `Follow` model (`follower`, `followed`, `created_at`); add a `UniqueConstraint` on `(follower, followed)` and prevent self-follow at the model/service level
- Generate and apply the migration
- `POST /users/:id/follow` — authenticated users only; idempotent (re-following returns 200, not 400); returns 404 if target user not found; returns 400 on self-follow attempt
- `DELETE /users/:id/follow` — authenticated users only; idempotent (unfollowing a non-followed user returns 200); returns 404 if target user not found
- Emit a `new_follower` event stub on successful follow (no-op until Week 3 wires the notification trigger)
- Add tests covering: follow, unfollow, re-follow idempotency, self-follow prevention, unauthenticated access

## Acceptance Criteria
- [ ] `Follow` model exists with the correct fields and unique constraint
- [ ] Migration is generated as a new file (not a modification to an existing migration)
- [ ] `POST /users/:id/follow` creates a follow record; re-follow is idempotent
- [ ] `DELETE /users/:id/follow` removes the follow record; unfollow of non-followed user is idempotent
- [ ] Self-follow returns 400 with a descriptive error
- [ ] Both endpoints return 401 for unauthenticated requests
- [ ] `new_follower` event is emitted on successful follow
- [ ] All happy-path and error-path cases covered by automated tests

Effort: M
Deadline: 2026-04-20

KmalM:
# 📝 Description
Fixes #399

Adds the core follow/unfollow write-side for the social graph feature.

**Changes:**
- `Follow` model with `UniqueConstraint(follower, followed)` and `CASCADE` on both FK sides
- Migration `0004_follow.py`
- `NEW_FOLLOWER` added to `NotificationType`
- `signals.py` stub — `on_follow_created` no-op, wired via `apps.py` ready()
- `follow_user` service: `get_or_create` wrapped in `transaction.atomic()` for MySQL safety; idempotent; self-follow → 400; inactive/missing target → 404
- `unfollow_user` service: idempotent delete; missing target → 404
- `FollowView` at `POST/DELETE /users/<id>/follow/`: 201 on first follow, 200 on re-follow, 204 on unfollow

# 📊 Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Documentation

# ⏱️ Estimated Review Time
- [ ] < 5 min
- [x] 5-15 min
- [ ] > 15 min


AysuKeskin (changes requested):
Overall, this looks solid. The follow/unfollow flow is implemented cleanly across model, service, view, serializer, migration, and tests, and I especially liked the use of `transaction.atomic()` together with the `IntegrityError` fallback for MySQL compatibility. The idempotent follow/unfollow behavior is clear, fixture usage is consistent, and the test coverage is in good shape, including the main success and error cases as well as cascade delete behavior.

I noted a few things to address before approval:

1. In `test_services.py`, the new imports for `Follow`, `follow_user`, and `unfollow_user` should be moved to the top of the file with the other imports. Right now they are added in the middle/bottom of the file, which goes against normal import organization and PEP 8 expectations.

2. There is an inconsistency between `follow_user` and `unfollow_user` around `is_active`. `follow_user` only allows following active users, but `unfollow_user` checks only whether the target user exists. That makes the behavior asymmetric. If this is intentional, it would be good to document the reasoning; otherwise, I think `unfollow_user` should be aligned with the same active-user rule.

3. `FollowUserSerializer` works as written, but I think this would be a better fit for a `ModelSerializer`. Since the current version relies on `follower_id` and `followed_id` implicitly matching Django’s FK naming convention, using `ModelSerializer` would make it a bit more explicit and less fragile.

Once those are cleaned up, this should be in a good state.

KmalM (commented):
Fixed items 1 and 3. On item 2, I'm keeping the behavior but added a doc comment explaining the reasoning.

**1. Import organization** — moved `Follow`, `follow_user`, and `unfollow_user` to the top-level imports in `test_services.py`, removed the mid-file import block.

**2. is_active asymmetry** — keeping it intentionally asymmetric. If `unfollow_user` also checked `is_active`, a deactivated user's followers would have no way to clean up the stale follow relationship (unfollow would 404). Added a doc comment to make the reasoning explicit.

**3. FollowUserSerializer → ModelSerializer** — converted. The `Meta` class now references the `Follow` model directly, which makes the field binding explicit and fails loudly at import if the FK names ever change.

All 268 tests pass.

AysuKeskin (approved):
Issues I previously requested changes on now seem to be resolved, and all tests pass. Ready to merge!

PR #421, opened 2026-04-19
feat: follower/following list endpoints and public profile counts

### LINKED ISSUE #400: [Task][Backend] Implement followers/following list endpoints and profile counts
## Description
Implement the read-side of the follow/unfollow social graph: paginated follower/following lists and counts in the public profile response.

Depends on #399 (Follow model and schema must exist first).

Tasks:
- `GET /users/:id/followers` — returns paginated list of users following the given user; accessible to all (authenticated or not)
- `GET /users/:id/following` — returns paginated list of users the given user follows; accessible to all (authenticated or not)
- Add `followers_count` and `following_count` fields to the public user profile response (`GET /users/:id`)
- Add tests covering: pagination, correct counts after follow/unfollow, 404 on unknown user

## Acceptance Criteria
- [ ] `GET /users/:id/followers` returns a paginated list with correct entries
- [ ] `GET /users/:id/following` returns a paginated list with correct entries
- [ ] Both list endpoints return 404 when the target user does not exist
- [ ] Public profile response includes `followers_count` and `following_count` that reflect the current state
- [ ] Counts update correctly after a follow or unfollow action
- [ ] All cases covered by automated tests

Effort: M
Deadline: 2026-04-20

KmalM:
# 📝 Description
Fixes #400

Adds follower and following list endpoints, and exposes `followers_count` / `following_count` on the public profile endpoint. Builds on the `Follow` model introduced in #420.

**New endpoints:**
- `GET /users/<id>/followers/` — paginated list of users who follow the given user
- `GET /users/<id>/following/` — paginated list of users the given user follows

**Updated endpoint:**
- `GET /users/<id>/` now includes `followers_count` and `following_count` integer fields

**Key implementation details:**
- `FollowedUserSerializer` applies the same username/photo privacy rules as `PublicUserProfileSerializer`
- `FollowerListView` and `FollowingListView` use `StoryPagination` (existing `PageNumberPagination` class)
- `get_public_profile` uses `Count('followers', distinct=True)` / `Count('following', distinct=True)` annotations; `distinct=True` is required on all three `Count` calls to avoid row inflation from cross-joins
- Returns 404 when the target user does not exist or is inactive

# 📊 Type of Change
- [x] New feature

- [x] New and existing unit tests pass locally with my changes (291 passed)

# 📸 Screenshots / Recordings

# ⏱️ Estimated Review Time
- [ ] < 5 min
- [x] 5-15 min
- [ ] > 15 min


AysuKeskin (changes requested):
Overall, this looks good. The follower/following list endpoints fit the existing structure well, and I especially liked the use of `select_related('profile')`, the `distinct=True` handling for count annotations, the privacy-aware serializer behavior, and the fact that the list endpoints are kept public with pagination consistent with the rest of the project.

I have one important comment and a few smaller test-related suggestions before approval.

The main issue is that the queryset construction and the `404` existence checks are currently implemented directly in the views. Since the project structure expects views to stay thin and delegate business/query logic to the service layer, I think this should be moved into `services.py` as dedicated functions such as follower/following list retrieval helpers. This is the only point I’d consider important, because it affects the code structure and separation of responsibilities rather than just test quality.

The remaining comments are more minor and mostly related to tests, but they would still be worth cleaning up:

* In both `TestFollowerListView` and `TestFollowingListView`, `test_returns_200` and `test_unauthenticated_returns_200` appear to be checking the same thing, since the `client` fixture is already unauthenticated. One of each pair could be removed to avoid duplication.

* I could not find a test covering the inactive-user path, even though both views explicitly return `404` when `is_active=False`. There is a `test_unknown_user_returns_404`, but it would be good to also have a dedicated inactive-user `404` case in both list view test classes.
* The serializer seems to include visibility logic for cases like hidden usernames and hidden profile photos, but I do not see direct tests for those branches in the follower/following list response tests. Since that privacy behavior is one of the more meaningful parts of the serializer, adding explicit tests for `is_username_public=False` and `is_photo_public=False` would make the coverage more complete.
* In `test_followers_count_decrements_after_unfollow`, the setup/removal currently goes through the ORM directly instead of the unfollow service. Since the rest of the follow/unfollow flow is mostly exercised through service functions, I think using `unfollow_user()` there as well would make the test more consistent with the intended abstraction.
* In several tests such as `test_lists_correct_followers` / `test_lists_correct_following` and similar follow-list tests, `Follow` is imported inline inside the test method. I think that import should be moved once to the top of the file with the other imports.

None of these are major blockers on their own, but addressing them would make the test suite cleaner and more complete. Once the service-layer refactor is done and these test issues are cleaned up, I think the PR will be in a strong state.

AysuKeskin (changes requested):
This is in a better place now.

What looks good:
- The main structural concern from the previous review has been addressed well.
- The follower/following queryset construction and the target-user existence checks have been moved out of the views and into the service layer.
- As a result, the views are now much thinner and more consistent with the overall project structure.
- I also confirmed that inactive-user `404` cases were added for both list endpoints, which closes one of the clearer test gaps from the previous round.

What still needs to be addressed before approval:
- I still see one important behavioral issue: inactive or soft-deactivated users can still leak into public follow surfaces.
- `get_followers()` and `get_following()` do not filter related users by `is_active=True`.
- The `followers_count` / `following_count` annotations in `get_public_profile()` also appear to count follow relations without excluding inactive users.
- That means a deactivated account may still:
  - appear in `/users/:id/followers/`
  - appear in `/users/:id/following/`
  - continue inflating the public follow counts on the profile endpoint

Why I think this matters:
- These are public-facing endpoints and counts.
- Because of that, inactive users should be handled consistently.
- I think the fix should be one of these:
  - exclude inactive users from the follower/following querysets and counts, or
  - clean up follow edges during deactivation

KmalM:
Addressed the feedback from the last review. Here's what changed:

**`services.py`**
- `get_followers()` and `get_following()` now filter the returned queryset by `is_active=True`, so deactivated accounts no longer appear in public follower/following lists.
- `get_public_profile()`: both `followers_count` and `following_count` annotations now carry a `filter=Q(...)` clause (`followers__follower__is_active=True` and `following__followed__is_active=True`) so inactive users are not counted on the public profile either.

**Tests**
- Added `test_inactive_follower_excluded_from_list` and `test_inactive_following_excluded_from_list` to the view tests for both list endpoints.
- Added `test_followers_count_excludes_inactive_follower` and `test_following_count_excludes_inactive_followed` to both the service-level and view-level test classes for follow counts.
- Moved the inline `from apps.users.models import Follow` imports that were scattered inside individual test methods up to the top-level import in `test_views.py`.

All 196 tests pass.

AysuKeskin (approved):
The main behavioral issue from the previous round is fully resolved — inactive     
 users are correctly excluded from all three public surfaces: the followers list,   
 the following list, and the profile count annotations. The fix is applied 
 consistently and the tests confirm it at both the service and view layers.
Approved.

PR #424, opened 2026-04-20
fix: indicate search results clearly in map view

### LINKED ISSUE #405: [Bug][Frontend] Implement Search Result Status Indicators on Map Interface
### Description

Currently, the map interface lacks clear visual feedback regarding the number of stories found after a user performs a search or pans to a new area. While the Feed View already provides "X stories found" or "No stories found" signals, the Map View remains silent, leaving users to guess if the lack of pins is due to a slow connection, a system error, or an actual lack of content in that specific location (e.g., Beykoz).

To improve UX consistency and reduce user frustration, we need to bring the search status indicators from the Feed into the Map UI. When a user searches for a location or filters content, a floating status badge or "snackbar" should appear on the map to provide immediate confirmation of the results.

### Acceptance Criteria

- [ ] A status indicator (e.g., a pill-shaped badge or top-aligned overlay) must appear on the Map interface following a search or area update.
- [ ] If stories exist, display the count: "3 stories found in this area".
- [ ] If no stories exist, display: "No stories found with this criteria" or "No stories found in [Location Name]".
- [ ] The indicator should update dynamically as the user triggers a new search, or disappear if the search is cleared.
- [ ] The styling (font, colors, and phrasing) must match the existing indicators used in the Feed View.
- [ ] Unit tests are written or updated.

### Meta (optional)

_No response_

sezindogan:
# 📝 Description
Fixes #405 
Adds a search result status indicator to the Map View. After a search or filter is applied, a small text badge appears inside the search panel showing how many stories were found (or a "no results" message). The indicator also notes when results are capped at 100. Matches the phrasing and styling used in the Feed View. 

### Changes Made
- src/services/storyService.js: getMapStories now returns a totalCount field alongside the FeatureCollection: the raw count from the search API (so we can detect the 100-result cap) or features.length for the map-endpoint path.

- src/pages/MapPage.jsx: Added hasActiveFilters from useFilterState, a totalCount state, and a status indicator paragraph rendered inside the search panel overlay. It shows "X stories/story found", "No stories found match your search.", and an extra "Showing first 100 results." line when q is set and totalCount > 100.

- Tests: Updated storyService.test.js to assert on the new totalCount field, and added 6 new test cases to MapPage.test.jsx covering count display, singular/plural, empty state, cap notice, loading suppression, and no-filter suppression.

# 📊 Type of Change
- [x]  Bug fix
- [ ]  New feature
- [ ]  Refactoring
- [ ]  Documentation
 
# 📸 Screenshots / Recordings
Screenshot 2026-04-20 at 16 38 52

# ⏱️ Estimated Review Time
- [x] < 5 min
- [ ] 5-15 min
- [ ] > 15 min

### DIFF
--- a/frontend/src/pages/MapPage.jsx
+++ b/frontend/src/pages/MapPage.jsx
@@ -8,9 +8,10 @@ import { useFilterState } from "@/hooks/useFilterState";
 const EMPTY_FEATURE_COLLECTION = { type: "FeatureCollection", features: [] };
 
 function MapPage() {
-  const { q, yearFrom, yearTo, location } = useFilterState();
+  const { q, yearFrom, yearTo, location, hasActiveFilters } = useFilterState();
 
   const [featureCollection, setFeatureCollection] = useState(EMPTY_FEATURE_COLLECTION);
+  const [totalCount, setTotalCount] = useState(0);
   const [loading, setLoading] = useState(true);
   const [error, setError] = useState(null);
 
@@ -19,7 +20,9 @@ function MapPage() {
     setError(null);
     try {
       const data = await getMapStories({ q, yearFrom, yearTo, location });
-      setFeatureCollection(data ?? EMPTY_FEATURE_COLLECTION);
+      const fc = data ?? EMPTY_FEATURE_COLLECTION;
+      setFeatureCollection(fc);
+      setTotalCount(fc.totalCount ?? fc.features?.length ?? 0);
     } catch (err) {
       setError(
         err?.response?.data?.detail ||
@@ -43,6 +46,16 @@ function MapPage() {
       <div className="absolute top-3 left-3 right-3 z-[1000] sm:left-4 sm:right-auto sm:w-[32rem]">
         <div className="rounded-lg border bg-background/95 p-3 shadow-md backdrop-blur-sm">
           <SearchFilter />
+          {hasActiveFilters && !loading && !error && (
+            <p className="mt-2 text-sm text-muted-foreground" aria-live="polite">
+              {featureCollection.features.length === 0
+                ? "No stories match your search."
+                : `${featureCollection.features.length} ${featureCollection.features.length === 1 ? "story" : "stories"} found.`}
+              {featureCollection.features.length > 0 && q && totalCount > 100 && (
+                <span className="block text-xs">Showing first 100 results.</span>
+              )}
+            </p>
+          )}
         </div>
       </div>
 
--- a/frontend/src/pages/MapPage.jsx
+++ b/frontend/src/pages/MapPage.jsx
@@ -2,7 +2,7 @@ import { useState, useEffect, useCallback } from "react";
 
 import MapView from "@/components/MapView/MapView";
 import SearchFilter from "@/components/SearchFilter/SearchFilter";
-import { getMapStories } from "@/services/storyService";
+import { getMapStories, MAP_SEARCH_CAP } from "@/services/storyService";
 import { useFilterState } from "@/hooks/useFilterState";
 
 const EMPTY_FEATURE_COLLECTION = { type: "FeatureCollection", features: [] };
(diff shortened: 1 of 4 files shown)


omarreis01 (commented):
## PR Review: fix: indicate search results clearly in map view

### Summary
Small, focused UX fix that mirrors the Feed View's result-count phrasing inside the Map View's search panel. The service layer now surfaces a `totalCount` so the page can detect the 100-result cap on the search path. Nicely scoped change, tests added alongside.

---

### Frontend

**MapPage.jsx**
- Good: `aria-live="polite"` on the indicator — screen readers will announce updates without stealing focus. Matches feed behavior.
- Minor: `totalCount` is tracked as a separate piece of state, but it always mirrors `featureCollection.totalCount`. You could drop the state and just read `featureCollection.totalCount ?? featureCollection.features.length ?? 0` directly in render. Not blocking — the explicit state is arguably clearer.
- Minor: The `100` cap threshold is duplicated between `MapPage.jsx` (line ~55) and `storyService.js` (line 72 — `page_size: 100`). If the cap ever changes, two places need updating. Consider exporting a `MAP_SEARCH_CAP` constant from the service. Non-blocking.
- Copy nit: the PR description says `"No stories found match your search."` but the code renders `"No stories match your search."`. The code is correct/natural — just flagging the mismatch.

**storyService.js**
- Clean refactor. The two code paths (`/search` vs `/map`) both now return a shape with `totalCount`, which is a nicer contract than before. The `/map` fallback to `features.length` is reasonable since that endpoint isn't paginated.

---

### Test coverage

6 new tests added — good coverage on the golden path, singular/plural, empty state, cap notice, and no-filter suppression. One concern:

- **`\"does not show indicator while loading\"`** uses `screen.queryByRole(\"paragraph\")`. `<p>` elements have no implicit ARIA role, so `queryByRole(\"paragraph\")` always returns `null` — the assertion passes trivially regardless of what the component renders. This is a false-positive test. Suggest replacing with a negative text match, e.g.:
  ```js
  expect(screen.queryByText(/stories? found|no stories match/i)).not.toBeInTheDocument();
  ```
  That would actually verify the indicator is suppressed during loading.

Otherwise the tests look solid.

---

This PR doesn't add any new endpoints — it only reads fields already returned by existing ones. Verified against `backend/apps/stories/views.py` + `common/pagination.py`:

- `GET /stories/search/` — ✅ `StorySearchView` uses `paginator.get_paginated_response(...)` which returns `{count, next, previous, results}`. Frontend reads `response.data.count` → matches.
- `GET /stories/map/` — ✅ Returns `{type: \"FeatureCollection\", features: [...]}` (no count). Frontend correctly falls back to `features.length`.
- `page_size: 100` — ✅ `StoryPagination.max_page_size = 100`, so requesting 100 is honored. The hard cap logic in the UI (`totalCount > 100`) correctly detects when the backend truncated.

No mismatches.

---

### Verdict
**APPROVE (with one test fix worth doing before merge)**

Non-blocking notes:
- Fix the false-positive `queryByRole(\"paragraph\")` test so the loading-suppression case is actually verified.
- Consider deduplicating the `100` cap constant between `storyService.js` and `MapPage.jsx`.
- Drop the redundant `totalCount` state if you agree it's cleaner to read straight from `featureCollection`.

Nice tight PR — the UX affordance matches the feed, and the service-layer change is the right shape. 👍

omarreis01 (approved):
Looks good — small non-blocking notes in the comment above (one false-positive test worth fixing).

sezindogan:
Thank you for the detailed review. I have applied all the suggested fixes:
  - storyService.js: Exported MAP_SEARCH_CAP = 100 constant and used it for the page_size param, replacing the magic 100.                              
  - MapPage.jsx: Imported MAP_SEARCH_CAP; dropped the redundant totalCount state (now reads featureCollection.totalCount directly in render); uses MAP_SEARCH_CAP in the cap threshold check.                                                                                                           
  - MapPage.test.jsx: Fixed the false-positive loading test (replaced queryByRole("paragraph") with a real text matcher); added MAP_SEARCH_CAP: 100 to the service mock so the constant is available in tests.

PR #430, opened 2026-04-21
added sorting by like count to get_story_feed()

### LINKED ISSUE #429: [Task][Backend] Implement backend for sorting feed based on popularity
### Description

Modify backend function used by API's that return story. The modification should ensure most liked stories show up first.

### Acceptance Criteria

- [x] The function use by story getting API's use is modified. 
- [x] Unit tests are made

### Meta (optional)

_No response_

AHMET-CAGDAS-GIRIT:
# 📝 Description
Fixes #429
# 📊 Type of Change
- [ ]  Bug fix
- [X]  New feature
- [ ]  Refactoring
- [ ]  Documentation
 
- [X] Project builds successfully
- [X] My code follows the style guidelines of this project
- [X] I have commented my code, especially in hard-to-understand parts
- [X] I have performed a self-review of my code
- [X] I have added/updated tests
- [X] New and existing unit tests pass locally with my changes

# 📸 Screenshots / Recordings

# ⏱️ Estimated Review Time
- [X] < 5 min
- [ ] 5-15 min
- [ ] > 15 min

### DIFF
--- a/backend/apps/stories/services.py
+++ b/backend/apps/stories/services.py
@@ -60,8 +60,9 @@ def get_story_feed(sort_by='recent', year_from=None, year_to=None, location=None
 
     if sort_by == 'recent':
         qs = qs.order_by('-submitted_at')
-
-    # TODO: add sort_by='popular' ordered by like_count once interactions app is implemented
+        
+    if sort_by == 'popular':
+        qs = qs.order_by('-like_count')
 
     return qs
 
--- a/backend/apps/stories/services.py
+++ b/backend/apps/stories/services.py
@@ -38,8 +38,7 @@ def get_story_feed(sort_by='recent', year_from=None, year_to=None, location=None
     Location filtering is a case-insensitive substring match on location_name so
     that partial queries like "galata" match "Galata Bridge".
 
-    sort_by='popular' is reserved for when the interactions app provides like_count
-    data. Until then only 'recent' is supported.
+    sort_by — 'recent' (default) orders by submission date; 'popular' orders by like_count
     """
     qs = Story.objects.filter(status=Story.STATUS_PUBLISHED)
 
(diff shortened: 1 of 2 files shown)


AysuKeskin (changes requested):
The logic itself looks correct and I do not expect any runtime issues here, but I have two very small cleanup suggestions that I think would be worth addressing before merge.

1. `if` → `elif` in `services.py:64`

```python
# current
if sort_by == 'popular':

# suggested
elif sort_by == 'popular':
````

Since these conditions are mutually exclusive, `elif` is the better fit here. Using a second `if` is not functionally wrong in the current code, but it makes the control flow look less clear and may suggest that both branches could run. It would also make the chain easier to extend safely later.

2. Update the stale docstring in `services.py:41–42`

The current docstring still says that `sort_by='popular'` is reserved for future support, but that behavior is now implemented. So the comment no longer matches the actual code.

Something like this would be more accurate:

```python
sort_by — 'recent' (default) orders by submission date; 'popular' orders by like_count DESC.
```

Neither of these is a major issue, but both seem like quick fixes that would improve clarity before merge.

AHMET-CAGDAS-GIRIT:
Did the changes.

> The logic itself looks correct and I do not expect any runtime issues here, but I have two very small cleanup suggestions that I think would be worth addressing before merge.
> 
> 1. `if` → `elif` in `services.py:64`
> 
> ```python
> # current
> if sort_by == 'popular':
> 
> # suggested
> elif sort_by == 'popular':
> ```
> 
> Since these conditions are mutually exclusive, `elif` is the better fit here. Using a second `if` is not functionally wrong in the current code, but it makes the control flow look less clear and may suggest that both branches could run. It would also make the chain easier to extend safely later.
> 
> 2. Update the stale docstring in `services.py:41–42`
> 
> The current docstring still says that `sort_by='popular'` is reserved for future support, but that behavior is now implemented. So the comment no longer matches the actual code.
> 
> Something like this would be more accurate:
> 
> ```python
> sort_by — 'recent' (default) orders by submission date; 'popular' orders by like_count DESC.
> ```
> 
> Neither of these is a major issue, but both seem like quick fixes that would improve clarity before merge.

Did the changes.

AysuKeskin (approved):
It looks right, now. Feature is implemented correctly. Approving.

Issue #216, opened 2026-03-26
[Feature][Backend] Enhance tag-based filtering across feed and map endpoints


AysuKeskin:
## Description

The current story discovery functionality should be extended to better match the project requirements. At the moment, search and filtering are limited, but users are expected to discover stories not only by title and location name, but also by tags. In addition, tag filtering should work together with the existing time and location filters across both feed and map views.

The goal of this issue is to improve the existing backend search and filtering behavior rather than introduce a separate search system.

The feed and map endpoints should also accept a tag filter parameter and handle it correctly together with time and location filters. Combined filters should narrow results predictably, and feed and map results should remain consistent with the active search and filtering criteria.

This issue should include the required query updates, filter parameter handling, validation, and tests for both individual and combined search/filter scenarios.

### Expected Behavior

**Filter enhancement**
- Feed and map endpoints accept a `tag` filter parameter
- Tag filter works together with existing time and location filters
- Combined filters narrow results correctly
- Active search/filter criteria produce consistent results across endpoints

## Acceptance Criteria

- [ ] Feed endpoint accepts a `tag` filter parameter
- [ ] Map endpoint accepts a `tag` filter parameter
- [ ] Tag, time, and location filters can be combined correctly
- [ ] Search and filter behavior remains consistent across feed and map results
- [ ] Appropriate validation/error handling exists for unsupported or malformed filter input
- [ ] Tests are added for search by title, narrative text, tag, and location
- [ ] Tests are added for combined filtering scenarios
- [ ] New and existing relevant tests pass

## Meta (optional)

Deadline: 22-04-2026
Estimated Duration: 3 hours

Issue #212, opened 2026-03-26
[Feature][Backend] Implement tag page endpoint for related stories


AysuKeskin:
### Description

The requirements specify that tags should not only be assignable metadata, but also part of the story discovery experience. In particular, users should be able to click on a tag and view stories associated with that tag in a feed-style page. This means the backend must provide a dedicated endpoint that resolves a tag by slug and returns the related stories in a format that can be rendered similarly to the main feed.

The goal of this issue is to implement the following endpoint:

- `GET /tags/:slug/stories`

This endpoint should accept a tag slug, resolve the corresponding tag record, and return a paginated list of stories associated with that tag. The returned data should follow the same general structure and pagination conventions already used by the feed endpoint so that the frontend can render the results as a filtered feed without needing a completely separate integration pattern. Reusing the feed-like structure here is important for consistency across the story browsing experience.

The endpoint should also respect the platform’s story visibility rules. Only stories that are meant to be discoverable should appear in the result set. Removed, hidden, draft, or otherwise non-public stories should remain excluded according to the project’s story-publication logic. If the provided tag slug does not correspond to any existing tag, the endpoint should return a clear and appropriate error response rather than silently returning unrelated data.

This issue should cover the full backend implementation needed for this tag page experience, including the tag lookup logic, queryset construction through the story-tag relationship, pagination behavior, serializer or response formatting, API view implementation, URL registration, and tests. The tests should include not only successful retrieval, but also cases such as non-existent slugs and valid tags with no currently visible stories.

This issue is important because it turns tags from passive metadata into an actual navigation and discovery feature, which aligns directly with the project’s goal of helping users explore related stories by topic, place, and theme.

### Expected Behavior
- Endpoint: `GET /tags/:slug/stories`
- Accepts a tag slug in the URL
- Returns a paginated set of published stories associated with that tag
- Uses a feed-compatible response structure
- Returns an error if the tag does not exist
- Excludes removed or otherwise non-visible stories

### Acceptance Criteria

- [ ] `GET /tags/:slug/stories` is implemented
- [ ] The endpoint resolves tags by slug
- [ ] Returned results are paginated
- [ ] Returned story data is compatible with feed rendering
- [ ] Only published/visible stories are returned
- [ ] Non-existent tag slugs return an appropriate error response
- [ ] Tests are added for successful retrieval, missing tag, and empty-result cases
- [ ] New and existing relevant tests pass

### Meta (optional)

Deadline: 26-04-2026
[...]

PR #431, opened 2026-04-21
Feat/tag endpoints

### LINKED ISSUE #209: [Feature][Backend] Implement tag listing and creation endpoints
## Description

The platform’s story submission and discovery flows rely on a reusable tag system that allows stories to be categorized consistently and browsed more easily. According to the project requirements, users should be able to choose from existing tags and also create new ones when needed, as long as the resulting tag structure remains normalized and predictable. In particular, tags are expected to follow a consistent slug format and should not be duplicated due to case differences, formatting variations, or repeated creation attempts.

The goal of this issue is to add the core backend endpoints required to support tag reuse and controlled tag creation:

- `GET /tags`
- `POST /tags`

The `GET /tags` endpoint should return the set of available tags so that frontend clients can use them in story creation forms, story update forms, filter UIs, and autocomplete/tag-selection components. This endpoint is expected to function as the canonical backend source for listing tags, so the response format should be stable, clean, and suitable for repeated use across different parts of the application.

The `POST /tags` endpoint should support creation of a new tag only when a matching slug does not already exist. Since the requirements emphasize structured tagging, the endpoint should validate the submitted tag data carefully and enforce the expected slug rules, especially the lower-case, dash-separated format. The implementation should also prevent duplicate tag creation in cases where two inputs are semantically the same but formatted slightly differently. This is important for keeping the tag system clean and avoiding fragmentation of story categorization over time.

Although this issue does not directly implement story-to-tag assignment, it should establish the backend behavior needed for that assignment flow to work smoothly in story create and update endpoints. That means the tag endpoints should be built in a way that is easy to integrate with the explicit story-tag relationship already present in the data model. The work should include serializer validation, API views, URL registration, any necessary service-layer logic, and tests for both normal and failure scenarios.

Overall, this issue is intended to provide the foundational API support for a consistent tagging system that can later be reused by story management, search, filter, and tag-based browsing features.

### Expected Behavior

**List tags**
- Endpoint: `GET /tags`
- Returns all available tags
- Response format is suitable for tag selection/autocomplete use cases

**Create tag**
- Endpoint: `POST /tags`
- Creates a new tag if the slug does not already exist
- Rejects duplicate slugs
- Enforces lower-case, dash-separated slug format
- Returns the created tag in the response

## Acceptance Criteria

- [ ] `GET /tags` is implemented
- [ ] `POST /tags` is implemented
- [ ] Tag creation rejects duplicate slugs
- [ ] Tag formatting rules are validated correctly
- [ ] Response format is usable for tag selection/autocomplete
- [ ] Story create/update flows can reuse these tags for assignment
- [ ] Tests are added for successful listing and creation cases
- [ ] Tests are added for duplicate and invalid-format cases

## Meta (optional)

Deadline: 27-04-2026
Estimated Duration: 4 hours

AysuKeskin:
# 📝 Description
Fixes #209

This PR adds foundational tag management support to the backend by implementing three tag endpoints with clear access restrictions and reusable service-layer logic.

### Implemented endpoints

| Method | URL | Access | Behavior |
|---|---|---|---|
| GET | `/tags/` | Everyone | Lists tags, with optional `?is_predefined=` and `?q=` filters |
| POST | `/tags/` | Authenticated registered users | Creates a tag idempotently (`201` if newly created, `200` if it already exists); if the caller is an admin, the created tag is marked as `is_predefined=True` |
| DELETE | `/tags/<pk>/` | Admin only | Deletes the selected tag |

### Authentication / authorization behavior
This PR intentionally applies different permission rules per endpoint:
- **GET `/tags/`** is public so tags can be used freely in selection, filtering, and autocomplete flows
- **POST `/tags/`** requires an authenticated registered user, so anonymous users cannot create tags
- **DELETE `/tags/<pk>/`** is restricted to admins only, so tag removal remains an administrative action

### What changed
- Added `TagSerializer` with normalization and validation for tag names
- Added predefined tag list support and seed migration
- Added service-layer functions for listing, creating, normalizing, and deleting tags
- Added `TagListCreateView` and `TagDetailView`
- Registered `/tags/` and `/tags/<pk>/` URL routes
- Implemented idempotent tag creation behavior
- Added filtering support for:
  - `?is_predefined=true/false`
  - `?q=<search>`
- Added test coverage for model, serializer, service, and view behavior

### Notes
- `POST /tags/` is idempotent by design:
  - returns **201 Created** when a new tag is created
  - returns **200 OK** when the normalized tag already exists
- The `is_predefined` field is not client-controlled through the request body
  - for regular users, created tags remain `is_predefined=False`
  - for admins, newly created tags are marked as `is_predefined=True`

# 📊 Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Documentation
 
# ⏱️ Estimated Review Time
- [ ] < 5 min
- [x] 5-15 min
- [ ] > 15 min

### DIFF
--- a/backend/apps/tags/admin.py
+++ b/backend/apps/tags/admin.py
@@ -1,1 +1,12 @@
-# admin.py — tags
+from django.contrib import admin
+
+from .models import Tag
+
+
+@admin.register(Tag)
+class TagAdmin(admin.ModelAdmin):
+    """Admin view for Tag — supports search and filtering by predefined status."""
+
+    list_display = ['name', 'is_predefined', 'story_count']
+    list_filter = ['is_predefined']
+    search_fields = ['name']
--- a/backend/apps/tags/migrations/0002_seed_predefined_tags.py
+++ b/backend/apps/tags/migrations/0002_seed_predefined_tags.py
@@ -1,0 +1,27 @@
+from django.db import migrations
+
+from apps.tags.predefined import PREDEFINED_TAGS
+
+
+def seed_predefined_tags(apps, schema_editor):
+    Tag = apps.get_model('tags', 'Tag')
+    for name in PREDEFINED_TAGS:
+        Tag.objects.update_or_create(name=name, defaults={'is_predefined': True})
+
+
+def unseed_predefined_tags(apps, schema_editor):
+    # On reverse: mark these tags as no longer predefined rather than deleting them,
+    # since user-created stories may already reference them via StoryTag.
+    Tag = apps.get_model('tags', 'Tag')
+    Tag.objects.filter(name__in=PREDEFINED_TAGS).update(is_predefined=False)
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('tags', '0001_initial'),
+    ]
+
+    operations = [
+        migrations.RunPython(seed_predefined_tags, reverse_code=unseed_predefined_tags),
+    ]
--- a/backend/apps/tags/predefined.py
+++ b/backend/apps/tags/predefined.py
@@ -1,0 +1,12 @@
+PREDEFINED_TAGS = [
+    'architecture',
+    'art',
+    'childhood',
+    'culture',
+    'daily-life',
+    'religion',
+    'war',
+    'food',
+    'natural-disasters',
+    'history',
+]
--- a/backend/apps/tags/serializers.py
+++ b/backend/apps/tags/serializers.py
@@ -1,23 +1,8 @@
-import re
-
+from django.core.exceptions import ValidationError as DjangoValidationError
 from rest_framework import serializers
 
 from .models import Tag
-
-
-def _normalize_name(value: str) -> str:
-    """
-    Normalize a raw tag input into lowercase-with-hyphens slug format.
-    Spaces and underscores become hyphens; non-alphanumeric characters are
-    dropped; consecutive hyphens are collapsed; leading/trailing hyphens
-    are stripped.
-    """
-    value = value.strip().lower()
-    value = re.sub(r'[\s_]+', '-', value)
-    value = re.sub(r'[^a-z0-9\-]', '', value)
-    value = re.sub(r'-{2,}', '-', value)
-    value = value.strip('-')
-    return value
+from .services import normalize_tag_name
 
 
 class TagSerializer(serializers.ModelSerializer):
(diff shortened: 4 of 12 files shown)


KmalM (changes requested):
Overall this is a clean, well-tested implementation of the tag listing and creation endpoints. The normalization pipeline, idempotent creation, and permission split are all done correctly. Two things need to be addressed before merging.

---

### 1. Story-tag attachment is not implemented — and has no tracking issue

The [Final Milestone Plan](https://github.com/bounswe/bounswe2026group4/wiki/Final-Milestone-Plan) Week 2 item explicitly says:

> **Tag endpoints** — `GET /tags` lists all tags; `POST /tags` creates a new tag if the slug does not exist; **assignment handled on story create/update**

This PR delivers the first half (listing + standalone creation) but not the attachment part. More importantly, there is no open issue tracking the story-tag assignment on story create/update. Without it:

- Users cannot actually tag a story via the API
- Downstream work (issue #212 tag page, issue #216 tag-based feed/map filtering) is blocked
- `story_count` will remain 0 for all tags indefinitely (see below)

Please open a follow-up issue for story-tag assignment (attach/detach tags on story create/update, enforce the 3-tag max, keep `story_count` in sync) before or alongside this PR so it does not fall through the cracks.

---

### 2. `story_count` is never incremented — the ordering is currently a no-op

`list_tags` orders by `-story_count, name`, but `story_count` has no update logic wired anywhere. There are no signals on `StoryTag` creation/deletion (unlike `Like` and `SavedStory` which have `post_save`/`post_delete` signals keeping `like_count`/`save_count` in sync). Until story-tag assignment lands, every tag will have `story_count=0`, meaning the ordering degrades to alphabetical only.

This is not a blocker in isolation — it is a consequence of the missing attachment work above. But it should be called out explicitly in the tracking issue so it is not forgotten.

AysuKeskin:
Thank you for the suggestion. I agree that this is an important missing part of the overall tag flow, and I have created a follow-up issue ( #432 ) to track it.

I think it is more appropriate to handle the feature you mentioned in a separate issue and PR, since it falls outside the current scope of this PR and requires a larger and more careful implementation. Because of that, I believe tracking and implementing it separately will make the work clearer, safer, and easier to review.

KmalM (changes requested):
Alright 👍

KmalM (approved):

PR #439, opened 2026-04-23
Feat/story tag assignment

### LINKED ISSUE #432: [Feature][Backend] Add story-tag assignment support on story create/update
### Description

The current tag implementation covers tag listing, standalone creation, and deletion, but story-tag assignment is still missing from the story create/update flow.

To make tags fully usable in practice, the backend should support attaching tags when a story is created and updating those associations when a story is edited. This includes handling both attach and detach behavior during story updates, enforcing validation rules at the serializer layer, and keeping tag-related counters synchronized correctly.

This work affects not only the tags app but also the story creation and update flow, especially in `apps/stories`, including serializers, services, and related request handling. It also introduces additional logic such as enforcing the maximum tag limit per story, preventing duplicate tags on a single story, and maintaining accurate `story_count` values for tags.

At the moment, `story_count` is used in ordering, but there is no logic yet to keep it synchronized with actual story-tag relations. That means tag ordering cannot reflect real usage until story-tag assignment and counter synchronization are implemented together.

This task should cover the missing backend behavior needed to support full story-tag assignment and make downstream tag-related features work correctly, including tag pages and tag-based feed/map filtering.

**Scope of this issue:**

* add tag attachment support when creating a story,
* add tag attach/detach update logic when editing a story,
* enforce the maximum 3-tag rule at the serializer/validation layer,
* prevent duplicate tags on a single story,
* keep `story_count` synchronized with story-tag associations,
* ensure this behavior is covered by backend tests.

This issue should also serve as the dependency-tracking item for the related downstream work on:

* tag page support,
* tag-based feed/map filtering.

### Acceptance Criteria

* [ ] Story create flow supports attaching tags
* [ ] Story update flow supports attaching and detaching tags
* [ ] A maximum of 3 tags per story is enforced
* [ ] Duplicate tags on the same story are prevented
* [ ] `story_count` is updated when story-tag relations are created or removed
* [ ] Signal or equivalent synchronization logic is added for `story_count`
* [ ] Tests are added or updated for create/update tag assignment behavior
* [ ] Tests are added or updated for `story_count` synchronization
* [ ] Related downstream work can reference this issue as a dependency

### Meta (optional)

Estimated Duration: 3–5 hours

AysuKeskin:
# 📝 Description
Fixes #432 

This PR completes the missing story-tag assignment flow by wiring tag attachment into story create/update operations and keeping `story_count` in sync.

### What changed
- Added `tags` ManyToMany support on `Story` through the existing `tags.StoryTag` table
- Added `tag_ids` as a write-only field on `StorySerializer`
- Added `tags` as a read-only field on story responses
- Validated tag assignment rules in `StorySerializer`:
  - maximum of 3 tags per story
  - no duplicate tag IDs
  - all referenced tag IDs must exist
- Updated `create_story(...)` to attach tags during story creation
- Updated `update_story(...)` to sync tags during story update
- Added tag attachment/sync service functions in `apps.tags.services`
- Added signals in `apps.tags.signals` to increment/decrement `story_count` on `StoryTag` create/delete
- Updated story list/detail querysets to prefetch tags
- Added migration to register the `Story.tags` relationship

### Story tag behavior
- On story create:
  - `tag_ids` may be omitted
  - if provided, the listed tags are attached to the new story
- On story update:
  - if `tag_ids` is omitted, existing tags remain unchanged
  - if `tag_ids` is provided, the story’s tag set is synced to that exact list
  - if `tag_ids` is an empty list, all tags are removed
- Story responses now include the current `tags` field in read format

### story_count behavior
This PR also addresses the previously missing `story_count` synchronization:
- creating a `StoryTag` increments the related tag’s `story_count`
- deleting a `StoryTag` decrements it
- syncing tags during story update correctly adjusts counts
- deleting stories also results in correct `story_count` cleanup through signal-driven `StoryTag` deletion behavior

### Tests added
This PR includes serializer, service, signal, and view-level tests covering:
- valid and invalid `tag_ids`
- max-3-tag enforcement
- duplicate and nonexistent tag rejection
- create/update attachment behavior
- no-change behavior when `tag_ids` is omitted in PATCH
- clearing all tags with empty `tag_ids`
- `tags` presence in story responses
- `story_count` increment/decrement behavior across attach, sync, and delete flows

# 📊 Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Documentation
 
# ⏱️ Estimated Review Time
- [ ] < 5 min
- [x] 5-15 min
- [ ] > 15 min


KmalM (changes requested):
## Race condition in `attach_tags_to_story` and `sync_story_tags` — `transaction.atomic()` missing

Both functions use a read-then-create pattern without a savepoint, which is unsafe under concurrent requests.

**`attach_tags_to_story`:**
```python
existing_ids = set(story.story_tags.values_list('tag_id', flat=True))
for tag_id in tag_ids:
    if tag_id not in existing_ids:
        StoryTag.objects.create(story=story, tag_id=tag_id)
```

**`sync_story_tags`:**
```python
for tag_id in to_add:
    StoryTag.objects.create(story=story, tag_id=tag_id)
```

Two concurrent PATCH requests for the same story with overlapping `tag_ids` can both pass the `existing_ids` check, then race to `CREATE`. One will hit the `unique_story_tag` constraint and raise `IntegrityError`. In MySQL, that poisons the entire outer transaction — any subsequent query in the same request will then raise `TransactionManagementError`. This is exactly the pattern called out in CLAUDE.md under "MySQL-specific gotcha: IntegrityError in tests".

**Fix** — wrap each create in a savepoint:

```python
from django.db import IntegrityError, transaction

try:
    with transaction.atomic():
        StoryTag.objects.create(story=story, tag_id=tag_id)
except IntegrityError:
    pass  # already linked — idempotent, safe to skip
```

This applies to the loop in `attach_tags_to_story` and to the `to_add` loop in `sync_story_tags`. The `transaction.atomic()` rolls back only to the savepoint on conflict, leaving the outer transaction intact.

AysuKeskin:
You're right, good catch. 
Both functions had the same read-then-create race that already burned us in the other flows. Fixed by wrapping each StoryTag.objects.create in a transaction.atomic() savepoint and catching IntegrityError. Also moved the StoryTag import to module level as a cleanup. Tests added for both functions to cover the IntegrityError path.
Ready for re-review.

KmalM (approved):
The race condition is fixed. Both `attach_tags_to_story` and `sync_story_tags` now wrap each `StoryTag.objects.create()` in `transaction.atomic()` with an `IntegrityError` catch, which is exactly the savepoint pattern needed to keep the outer MySQL transaction intact on a concurrent duplicate. The new mock-based tests for the `IntegrityError` path are a nice addition. LGTM.

PR #446, opened 2026-04-24
feat: add GET /users/:id/bookmarks/ saved stories endpoint

### LINKED ISSUE #219: [Feature][Backend] Implement saved stories endpoint for user bookmarks
### Description

In addition to allowing users to bookmark stories, the backend should provide a way to retrieve the stories that a user has saved. According to the requirements, the system should expose a saved stories endpoint that returns a user’s bookmarked stories in reverse chronological order. This will support profile-based saved content views and allow users to revisit stories they previously marked for later reading.

The goal of this issue is to implement:

- `GET /users/:id/bookmarks`

This endpoint should return the stories bookmarked by the specified user, ordered from most recently saved to oldest. The response should be suitable for rendering as a feed or list of saved stories on the frontend. The implementation should define and document whether this endpoint is only accessible by the owner of the bookmarks or whether it can also be accessed by admins if needed, but at minimum it should respect the project’s authentication and privacy expectations.

The endpoint should use the bookmark relationship rather than story creation or update timestamps to determine ordering. It should also exclude stories that are no longer visible or otherwise unavailable according to the system’s visibility rules. The implementation should include the necessary queryset logic, pagination if appropriate for consistency with other feed-like endpoints, serializer/view wiring, URL registration, permission handling, and tests.

### Expected Behavior
- Endpoint: `GET /users/:id/bookmarks`
- Returns the specified user’s bookmarked stories
- Results are ordered in reverse chronological order of bookmarking
- Response is suitable for frontend saved-story listing
- Non-visible or unavailable stories are handled appropriately
- Access follows the intended permission/privacy rules

### Acceptance Criteria

- [ ] `GET /users/:id/bookmarks` is implemented
- [ ] Results are ordered by most recently bookmarked first
- [ ] Bookmarked stories are returned for the correct user
- [ ] Non-visible or unavailable stories are handled correctly
- [ ] Permission/privacy behavior is enforced as intended
- [ ] Response format is suitable for saved story listing
- [ ] Tests are added for retrieval, ordering, empty results, and unauthorized/forbidden cases where applicable

### Meta (optional)

Deadline: 30-04-2026
Estimated Duration: 4 hours

KmalM:
# 📝 Description
Fixes #219

Implements `GET /users/:id/bookmarks/` — an owner-only paginated list of a user's saved stories, ordered by most recently bookmarked first.

**Behavior:**
- Returns `200` with a paginated `StoryFeedSerializer` response for the authenticated owner
- Returns `401` for unauthenticated requests
- Returns `403` for authenticated non-owners
- Returns `404` for nonexistent or inactive users
- Removed/draft stories are automatically excluded (only `STATUS_PUBLISHED` stories appear)
- `annotate_user_interactions` is applied before serialization to avoid N+1 queries on `user_has_liked` / `user_has_saved` fields

**Architecture:** Ownership enforcement lives in the service layer (`get_user_bookmarks` raises `PermissionDenied`). The view remains a thin orchestrator — service, annotate, paginate, serialize.

# 📊 Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Documentation

# 📸 Screenshots / Recordings
N/A — backend-only change

# ⏱️ Estimated Review Time
- [ ] < 5 min
- [x] 5-15 min
- [ ] > 15 min

### DIFF
--- a/backend/apps/stories/services.py
+++ b/backend/apps/stories/services.py
@@ -1,6 +1,21 @@
-from django.db.models import Q
+from django.db.models import Exists, OuterRef, Q
 
+from apps.interactions.models import Like, SavedStory
 from apps.stories.models import Story
+
+
+def annotate_user_interactions(qs, user):
+    """
+    Annotate a Story queryset with _user_has_liked and _user_has_saved boolean flags
+    for the given authenticated user.
+
+    Uses Exists subqueries so both checks are folded into the main SQL query,
+    avoiding N+1 when serializing paginated lists.
+    """
+    return qs.annotate(
+        _user_has_liked=Exists(Like.objects.filter(story=OuterRef('pk'), user=user)),
+        _user_has_saved=Exists(SavedStory.objects.filter(story=OuterRef('pk'), user=user)),
+    )
 
 
 def create_story(user, validated_data: dict) -> Story:
--- a/backend/apps/stories/views.py
+++ b/backend/apps/stories/views.py
@@ -1,29 +1,13 @@
-from django.db.models import Exists, OuterRef
 from rest_framework import generics
 from rest_framework.permissions import AllowAny, IsAuthenticated
 from rest_framework.response import Response
 from rest_framework.views import APIView
 
-from apps.interactions.models import Like, SavedStory
 from apps.stories.models import Story
 from apps.stories.serializers import FeedQuerySerializer, SearchQuerySerializer, StoryDetailSerializer, StoryFeedSerializer, StoryMapGeoJSONSerializer, StorySerializer
-from apps.stories.services import delete_story, get_story_feed, get_story_search
+from apps.stories.services import annotate_user_interactions, delete_story, get_story_feed, get_story_search
 from common.pagination import StoryPagination
 from common.permissions import IsOwnerOrAdmin
-
-
-def annotate_user_interactions(qs, user):
-    """
-    Annotate a Story queryset with _user_has_liked and _user_has_saved boolean flags
-    for the given authenticated user.
-
-    Uses Exists subqueries so both checks are folded into the main SQL query,
-    avoiding N+1 when serializing paginated lists.
-    """
-    return qs.annotate(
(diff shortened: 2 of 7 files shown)


AysuKeskin (changes requested):
Overall, the implementation looks solid and the bookmarks list functionality is well covered by both service-level and view-level tests.

I only have two issues that should be addressed before merge:

1. In `apps/users/views.py`, `annotate_user_interactions` is imported from `apps.stories.views`. I’m aware that this function is not introduced in this PR, but since it is being imported and used here, it still affects the layering. Importing from another app’s `views.py` breaks the expected architecture and can lead to tighter coupling over time. To avoid future layering issues, it would make sense to refactor this as part of this PR by moving `annotate_user_interactions` to `apps/stories/services.py` (or a more appropriate shared layer) and updating the import accordingly.

2. In the test files, there are inline imports inside test methods. Specifically, this appears in `apps/users/tests/test_services.py` and `apps/users/tests/test_views.py` for imports like `datetime` and `SavedStory`. These should be moved to the top of the files to keep the tests easier to scan and consistent with the project style.

After these two points are addressed, this should be good to merge.

KmalM:
Thanks for the review! Both changes are addressed in the latest push:

1. **`annotate_user_interactions` moved to `apps/stories/services.py`** — the function is no longer defined in `views.py`. Both `apps/stories/views.py` and `apps/users/views.py` now import it from `apps.stories.services`, which keeps the service layer as the correct cross-app boundary.

2. **Inline imports moved to file tops** — `datetime`, `timezone`, and `SavedStory` are now declared at the top of both `apps/users/tests/test_services.py` and `apps/users/tests/test_views.py`. No inline imports remain for these symbols.

Also rebased onto the latest `main` (picked up the follower/following routes from PR #421) — all 660 tests pass.

AysuKeskin (approved):
Thanks for addressing the requested changes.

I verified that `annotate_user_interactions` is now correctly moved to the service layer and that the cross-app import issue is resolved. The test files are also cleaned up with imports moved to the top, which improves readability and consistency.

There are no remaining blocking issues from my side. This is good to merge.

PR #447, opened 2026-04-24
fix: increase timeout for flaky RootNavigator auth test

### LINKED ISSUE #444: [Bug][Mobile] RootNavigator auth-flow test frequently times out in CI
## Description

The Mobile CI frequently fails on the same React Native navigation test:

`src/app/navigation/__tests__/RootNavigator.test.tsx`

Failing test:

`RootNavigator auth flow › shows a message instead of redirecting unauthenticated users for protected screens`

The failure is due to Jest’s default timeout being exceeded:

```txt
Exceeded timeout of 5000 ms for a test.
```

This does not happen on every run, but it has occurred repeatedly across multiple unrelated PRs. Because the same test fails intermittently in otherwise unrelated changes, this appears to be a **CI stability issue caused by a flaky test** rather than a regression introduced by a specific branch.

### Current Behavior

The test occasionally takes longer than Jest’s default 5000 ms timeout in the React Native CI environment. When that happens, Mobile Checks fail even though the rest of the mobile test suite passes.

### Expected Behavior

The `RootNavigator` unauthenticated protected-screen test should complete reliably in CI and should not randomly block unrelated PRs.

### Proposed Solution

Increase the timeout only for this specific test instead of changing the global Jest timeout.

This test renders the root navigator and waits for asynchronous auth/navigation behavior to settle, so it can reasonably take longer in CI than simpler component or unit tests. Scoping the timeout increase to this single test keeps the fix minimal and avoids masking slow behavior elsewhere in the suite.

Example:

```ts
it(
  'shows a message instead of redirecting unauthenticated users for protected screens',
  async () => {
    renderNavigator();

    await waitFor(() => {
      // existing expectation
    });
  },
  10000
);
```

The assertions themselves should remain unchanged. The goal is not to weaken the test, but to make it reliable under CI timing conditions.

## Acceptance Criteria

* [ ] Timeout is increased only for the flaky `RootNavigator` auth-flow test
* [ ] Existing test assertions remain unchanged
* [ ] No global Jest timeout change is introduced
* [ ] Mobile Checks pass more reliably in CI

## Meta (optional)

Estimated Duration: 1 hour

MertErenKaplan:
# 📝 Description
Fixes #444 

Increases the timeout only for the flaky `RootNavigator` auth-flow test in mobile so it does not intermittently fail in CI due to Jest's default 5000 ms limit.

# 📊 Type of Change
- [x]  Bug fix
- [ ]  New feature
- [ ]  Refactoring
- [ ]  Documentation
 
- [ ] Project builds successfully
- [ ] I have commented my code, especially in hard-to-understand parts
- [ ] I have added/updated tests

# 📸 Screenshots / Recordings
Not applicable.

# ⏱️ Estimated Review Time
- [x] < 5 min
- [ ] 5-15 min
- [ ] > 15 min

### DIFF
--- a/mobile/src/app/navigation/__tests__/RootNavigator.test.tsx
+++ b/mobile/src/app/navigation/__tests__/RootNavigator.test.tsx
@@ -280,7 +280,7 @@ describe('RootNavigator auth flow', () => {
 
     expect(await screen.findByText('Please sign in to submit a story.')).toBeTruthy();
     expect(screen.getByLabelText('Submission')).toBeTruthy();
-  });
+  }, 10000);
 
   it('opens the main pager on the feed tab by default', async () => {
     renderNavigator();


OsmanYusufTosun (approved):
Looks good to me. It increases the timeout only for the flaky RootNavigator auth-flow test, keeps the existing assertions intact. I also verified the targeted mobile test passes locally, so I’m approving.

PR #449, opened 2026-04-24
feat: add POST/DELETE /stories/:id/bookmark/ endpoints

### LINKED ISSUE #218: [Feature][Backend] Implement bookmark add/remove endpoints for stories
## Description

The platform should allow authenticated users to save stories for later viewing through a bookmark mechanism. According to the requirements, users should be able to bookmark a story and remove that bookmark later using dedicated endpoints. This functionality is intended to support a personalized reading and discovery experience without affecting the public story data itself.

The goal of this issue is to implement the following endpoints:

- `POST /stories/:id/bookmark`
- `DELETE /stories/:id/bookmark`

Both endpoints should be restricted to authenticated users. The create-bookmark action should save the target story for the current user, while the delete action should remove the saved relationship if it exists. The behavior should be idempotent, meaning repeated requests should not create duplicates or cause unexpected failures. For example, bookmarking an already bookmarked story should not create a second saved record, and deleting a bookmark that does not exist should still be handled safely according to the chosen API convention.

The implementation should ensure that bookmarks are associated with the authenticated user and the correct story, and that invalid story IDs are handled properly. It should also align with the project’s current interaction model and permission structure. This issue should include serializer or validation logic if needed, API views, URL registration, permission checks, and tests for successful, duplicate, missing, and unauthorized cases.

### Add bookmark
- Endpoint: `POST /stories/:id/bookmark`
- Requires authentication
- Saves the target story for the current user
- Does not create duplicate bookmark records

### Remove bookmark
- Endpoint: `DELETE /stories/:id/bookmark`
- Requires authentication
- Removes the bookmark for the current user if it exists
- Handles repeated delete requests safely

## Acceptance Criteria

- [ ] `POST /stories/:id/bookmark` is implemented
- [ ] `DELETE /stories/:id/bookmark` is implemented
- [ ] Only authenticated users can bookmark or unbookmark stories
- [ ] Bookmark creation is idempotent
- [ ] Bookmark deletion is idempotent or otherwise safely handled
- [ ] Invalid or non-existent story IDs return appropriate error responses
- [ ] Duplicate bookmark records are prevented
- [ ] Tests are added for successful, duplicate, missing, and unauthorized cases

## Meta (optional)

Deadline: 30-04-2026
Estimated Duration: 4 hours

KmalM:
# 📝 Description
Fixes #218

Implements `POST /stories/:id/bookmark/` and `DELETE /stories/:id/bookmark/` endpoints for adding and removing story bookmarks.

**Behavior:**
- `POST` returns `201` on first bookmark, `200` on a duplicate (idempotent — no second row created)
- `DELETE` always returns `204` regardless of whether a bookmark existed (idempotent)
- Both endpoints require a registered user (`IsRegisteredUser`)
- `Story.save_count` is kept in sync via Django post_save/post_delete signals (already wired to `SavedStory`)
- `transaction.atomic()` savepoint used on `get_or_create` to prevent MySQL `IntegrityError` from breaking the outer transaction

No migration required — the `SavedStory` model with its `UniqueConstraint` and signals already existed.

# 📊 Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Documentation

# 📸 Screenshots / Recordings
N/A — backend-only change

# ⏱️ Estimated Review Time
- [ ] < 5 min
- [x] 5-15 min
- [ ] > 15 min

### DIFF
--- a/backend/apps/interactions/serializers.py
+++ b/backend/apps/interactions/serializers.py
@@ -1,6 +1,6 @@
 from rest_framework import serializers
 
-from apps.interactions.models import Comment, Like
+from apps.interactions.models import Comment, Like, SavedStory
 
 
 class CommentCreateSerializer(serializers.Serializer):
@@ -33,3 +33,11 @@ class LikeResponseSerializer(serializers.ModelSerializer):
     class Meta:
         model = Like
         fields = ['id', 'story_id', 'created_at']
+
+
+class BookmarkResponseSerializer(serializers.ModelSerializer):
+    """Read-only representation of a SavedStory returned in API responses."""
+
+    class Meta:
+        model = SavedStory
+        fields = ['id', 'story_id', 'saved_at']
--- a/backend/apps/interactions/services.py
+++ b/backend/apps/interactions/services.py
@@ -116,16 +116,19 @@ def add_bookmark(user, story_id):
 
 def remove_bookmark(user, story_id):
     """
-    Delete the SavedStory for a published story if it exists.
+    Delete the SavedStory for a story if it exists.
 
-    Raises Http404 if the story does not exist or is not published.
+    Raises Http404 only if the story does not exist at all.
     Idempotent — no exception if the user has not bookmarked the story.
+
+    Status-agnostic by design: a user must be able to remove their own bookmark
+    even after a story is removed or unpublished.
 
     Uses instance.delete() (not queryset .filter().delete()) so that the post_delete
     signal fires and Story.save_count is decremented atomically.
     """
     try:
-        story = Story.objects.get(pk=story_id, status=Story.STATUS_PUBLISHED)
+        story = Story.objects.get(pk=story_id)
     except Story.DoesNotExist:
         raise Http404
 
--- a/backend/apps/interactions/services.py
+++ b/backend/apps/interactions/services.py
@@ -3,7 +3,7 @@ from django.http import Http404
 
 from rest_framework.exceptions import ValidationError
 
-from apps.interactions.models import Comment, Like
+from apps.interactions.models import Comment, Like, SavedStory
 from apps.stories.models import Story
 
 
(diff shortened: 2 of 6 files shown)


AysuKeskin (changes requested):
Overall, the implementation looks good, but there is one blocker that needs to be addressed before merge.

The issue is in the `remove_bookmark` logic (in `stories/services.py`). Currently, the removal operation applies a **published-only constraint**, which prevents users from removing bookmarks from stories that are no longer in a published state (e.g., removed/unpublished stories).

This creates an inconsistent and problematic behavior: users should be able to remove their bookmarks regardless of the current status of the story. Bookmark ownership is independent of whether the story is published or not.

For example, if a story gets removed after being bookmarked, the user is currently unable to clean up their own bookmarks, which is not desirable.

The removal behavior should instead be aligned with operations like `remove_like`, where the action depends on the relationship (user ↔ story), not on the story’s visibility/status.

In short:
- `remove_bookmark` should not filter by published status
- It should allow deletion as long as the bookmark relation exists

This should be fixed before approval.

KmalM:
Thanks for the review — you're absolutely right.

**Change made:** `remove_bookmark` no longer filters by `STATUS_PUBLISHED`. It now only raises `Http404` when the story doesn't exist at all (deleted from DB), not when it's unpublished or removed.

This aligns with the correct ownership semantics: a user should always be able to clean up their own bookmarks regardless of what happened to the story after they saved it.

**Tests updated:**
- `test_removed_story_raises_404` → replaced with `test_removed_story_can_be_unbookmarked` (verifies removal succeeds and the row is gone)
- `test_draft_story_raises_404` → replaced with `test_draft_story_can_be_unbookmarked` (same)
- `test_unbookmark_removed_story_returns_404` (view test) → renamed to `test_unbookmark_removed_story_returns_204` and asserts 204

AysuKeskin (approved):
Thanks for addressing the feedback.

I checked the updated implementation, and the `remove_bookmark` logic now correctly allows users to remove bookmarks regardless of the story’s published status, while still handling non-existent stories with `404` as expected.

The updated tests also reflect this behavior clearly and improve the overall consistency with similar operations like `remove_like`.

Everything looks good now, approving.

PR #454, opened 2026-04-26
Feat/notifications triggers and endpoints

### LINKED ISSUE #220: [Feature][Backend] Implement notification triggers and notification management endpoints
## Description

The platform requirements include a notification system that informs users about important events related to their stories, interactions, moderation outcomes, and achievements. In order to support this functionality, the backend should both generate notifications when specific trigger events occur and expose endpoints for retrieving and managing those notifications.

The goal of this issue is to implement backend support for notification creation on the following events:

- new comment on the user’s own story
- new like on the user’s own story
- moderation action on the user’s own content
- story removed
- report resolved
- new badge earned

In addition, the following endpoints should be implemented:

- `GET /notifications`
- `PATCH /notifications/:id/read`
- `DELETE /notifications`

The notification trigger side of this issue should ensure that notifications are created at the appropriate points in the relevant flows and are associated with the correct recipient. The system should avoid generating incorrect self-notifications where that would not make sense, and the notification payload should include enough information for the frontend to display the event meaningfully.

The endpoint side of this issue should allow authenticated users to retrieve their notifications, mark an individual notification as read, and clear all notifications when requested. `GET /notifications` should return only the notifications belonging to the authenticated user. `PATCH /notifications/:id/read` should update the read state of the specified notification if it belongs to the requester. `DELETE /notifications` should clear all notifications for the requester in a safe and predictable way.

This issue should include the trigger integration points, any service-layer notification creation helpers, endpoint implementation, permission checks, response design, and tests for both trigger generation and endpoint behavior.

#### Notification triggers
- Notifications are created for the required event types
- Notifications are linked to the correct recipient
- Notification content is suitable for frontend display
- Unwanted or incorrect notifications are avoided

#### Notification endpoints
- `GET /notifications` returns the authenticated user’s notifications
- `PATCH /notifications/:id/read` marks a single notification as read
- `DELETE /notifications` clears all notifications for the authenticated user

## Acceptance Criteria

- [ ] Notifications are emitted for new comment on own story
- [ ] Notifications are emitted for new like on own story
- [ ] Notifications are emitted for moderation action on own content
- [ ] Notifications are emitted for story removal
- [ ] Notifications are emitted for report resolution
- [ ] Notifications are emitted for new badge earned
- [ ] `GET /notifications` is implemented
- [ ] `PATCH /notifications/:id/read` is implemented
- [ ] `DELETE /notifications` is implemented
- [ ] Users can only access and modify their own notifications
- [ ] Tests are added for trigger generation and endpoint behavior

## Meta (optional)

Deadline: 07-05-2026
Estimated Duration: 5 hours

AysuKeskin:
# 📝 Description

Fixes #220

This PR implements the backend notification system.

It introduces a notification service layer, serializers, API views, URL routes, and signal-based triggers to automatically create notifications when relevant user actions occur. The supported notification events include:

- new comments on a story
- new likes on a story
- new followers
- badge earned
- new published stories (sent to followers)
- story removal by moderation
- report resolved or dismissed

The system is designed to be event-driven using Django signals, and all notification creation is centralized through a service function to keep the logic consistent and reusable.

The following API endpoints are added:

- `GET /notifications/` → list the authenticated user's notifications (unread first)
- `PATCH /notifications/:id/read/` → mark a notification as read/unread
- `DELETE /notifications/:id/` → delete a single notification
- `DELETE /notifications/` → clear all notifications

Notification creation avoids self-notifications where applicable (e.g. liking your own story). It also handles private usernames by displaying `"Someone"` and omitting the actor object to prevent leaking user identity.

Comprehensive tests are added for:
- service layer logic
- signal-based notification triggers
- API endpoints (including auth and ownership checks)
- edge cases such as private usernames and duplicate prevention

# 📊 Type of Change

- [x] New feature
- [ ] Bug fix
- [ ] Refactoring
- [ ] Documentation

- [x] Code builds successfully
- [x] Style guidelines are followed
- [x] Code has been self-reviewed
- [x] Tests have been added/updated
- [x] Local tests pass

# ⏱️ Estimated Review Time

- [ ] <5 min
- [ ] 5–15 min
- [x] >15 min


KmalM (changes requested):
Good overall implementation — the signal architecture is clean, the service layer is well-separated, and the test coverage is thorough. One real bug to fix before merging, plus a few worth discussing.

---

## 🐛 Bug — `story_id` / `comment_id` silently absent when null

**File:** `backend/apps/notifications/serializers.py`

```python
story_id = serializers.IntegerField(source='story.id', read_only=True, allow_null=True)
comment_id = serializers.IntegerField(source='comment.id', read_only=True, allow_null=True)
```

When `story` (or `comment`) is a null FK, `notification.story` returns `None`. DRF's attribute traversal then tries `getattr(None, 'id')`, which raises `AttributeError`. Because `read_only=True` makes `required=False`, DRF raises `SkipField` and the field is **omitted entirely** from the response — not serialized as `null`.

This means the test `test_list_response_contains_expected_fields` will fail: it creates a notification with no story/comment and then asserts `'story_id' in notif`.

**Fix:** access the raw FK column directly, which returns `None` when null:

```python
story_id = serializers.IntegerField(read_only=True, allow_null=True)   # uses story_id attr
comment_id = serializers.IntegerField(read_only=True, allow_null=True)  # uses comment_id attr
```

(DRF defaults `source` to the field name, so omitting `source` tells it to read `notification.story_id` — the integer column — rather than chaining `notification.story.id`.)

---

## ⚠️ Design gap — draft-to-published transitions don't notify followers

**File:** `backend/apps/stories/signals.py`

```python
@receiver(post_save, sender=Story)
def on_story_published(sender, instance, created, **kwargs):
    if not created or instance.status != Story.STATUS_PUBLISHED or not instance.user:
        return
```

`created=True` only fires on `INSERT`, not on `UPDATE`. If a user saves a draft and later publishes it, `created` is `False` and followers are never notified. The requirements don't say whether this case should notify, but it's worth explicitly documenting the decision (either add handling for status transitions or leave a comment saying it's intentional).

---

## ✅ What's good

- Signal handlers in each app's own `signals.py` + `AppConfig.ready()` keeps notification logic decoupled from views.
- `pre_save` with DB status comparison for `on_story_removed` and `on_report_resolved` correctly prevents duplicate notifications on subsequent saves.
- Private-username handling (actor set to None, message uses "Someone") is consistent across all signal handlers and tested.
- `create_notification` service is the single entry point for all notification creation — easy to extend later (rate limiting, async dispatch, etc.).
- Migration correctly widens the choices set; max_length=20 accommodates the longest value (`new_story_published`, 19 chars).

AysuKeskin:
Thanks for the review!

**Bug fix (`story_id` / `comment_id`):** You're right. I removed `source='story.id'` so DRF reads the raw FK integer column directly. Fixed and pushed.

**Draft → published transition:** I intentionally scoped this to `created=True` only. The app doesn't have draft functionality implemented or planned, so this case won't occur in practice ever. I decided not to add a comment for it to avoid documenting a scenario that isn't part of the current or future feature set.

KmalM (approved):
The serializer fix is correct — `story_id` and `comment_id` now read the raw FK integer columns directly, so they serialize as `null` (not absent) when the FK is null. The explanatory comment is a nice touch.

The two informational notes from the first review (draft-to-published gap in `on_story_published`, ordering vs. req 1.6.3.2) are documented in code and acceptable as-is. Good to merge.

PR #465, opened 2026-04-28
feat: integrate transactional email service for auth emails

### LINKED ISSUE #204: [Feature][Backend] Integrate transactional email service for auth emails
## Description

The authentication features require reliable outbound email support for user-facing flows such as email verification and password reset. In order for those features to function in a maintainable and production-ready way, the backend should be connected to a transactional email provider and should expose a reusable email-sending mechanism that auth flows can call without duplicating provider-specific logic. This issue covers that email service integration work.

The goal of this issue is to configure a transactional email provider, such as Resend or SendGrid, and connect it to the backend through a clean service abstraction. Rather than embedding provider-specific API calls directly into authentication views or services, the implementation should centralize email sending behind a dedicated module or helper so that verification and password reset flows can trigger emails in a consistent and testable way. This will also make future email-based features easier to add.

In addition to provider integration, the system should include the initial templates required for authentication-related emails. At minimum, this includes one template for the 6-digit verification code sent during registration and one template for the password reset link sent during account recovery. These templates should be clear, consistent, and easy to maintain, and the overall email configuration should be environment-based so that API keys, sender identity, and related settings are not hardcoded in the application.

This issue should also ensure that email sending failures are handled appropriately. The backend should fail gracefully, log relevant errors, and make the integration practical for both development and deployment environments. Developer setup documentation should be included so that team members know which environment variables and configuration values are required to run the email-related features locally and in deployed environments.

### Expected Behavior
- The backend can send transactional emails through a configured provider
- Email credentials and configuration are loaded from environment variables
- The system includes a reusable email sending service/module
- Verification code emails can be sent with the correct code content
- Password reset emails can be sent with a reset link
- Templates exist for both email types
- Failures are handled gracefully and logged appropriately

## Acceptance Criteria

- [ ] A transactional email provider is selected and configured
- [ ] Email configuration is managed through environment variables
- [ ] A reusable email service/module is added to the backend
- [ ] Verification code email template is implemented
- [ ] Password reset email template is implemented
- [ ] Auth flows can trigger email sending through the new service
- [ ] Error handling/logging is added for email send failures
- [ ] Required setup is documented for developers
- [ ] Relevant tests or mocked integration tests are added

## Meta (optional)

Deadline: 15-04-2026
Estimated duration: 3 hours

KmalM:
# 📝 Description
Fixes #204

Integrates a transactional email service for auth flows and adds a password reset endpoint pair. Also addresses 4 code quality issues found during review.

**Code quality fixes:**
- `LogoutView`: switched from `IsAuthenticated` to `IsRegisteredUser` to enforce the role check consistently with every other protected view
- `register_user`: wrapped `create_user` + `EmailVerificationCode.create` in `transaction.atomic()` to handle concurrent duplicate-registration race conditions in MySQL
- `CurrentUserProfileSerializer.birth_date`: added `allow_null=True` to prevent 500 errors for users with no birth date (the default state)
- `test_views.py`: removed local `registered_user` fixture that duplicated the shared `user` fixture from `conftest.py`

**Email service:**
- `apps/users/email_service.py` — `send_verification_email` and `send_password_reset_email` backed by Django's built-in mail system; failures are caught and logged, never propagated to callers
- Plain-text email templates for verification code and password reset link
- `PasswordResetToken` model with 1-hour expiry + migration
- `request_password_reset` service: silent for unknown/inactive emails (anti-enumeration — same response regardless of whether the email exists)
- `reset_password` service: validates token, updates password, marks token used
- `POST /auth/password-reset/` and `POST /auth/password-reset/confirm/` endpoints
- Email and `FRONTEND_URL` settings loaded from env vars; `.env.example` updated
- `register_user` now sends the real verification email instead of the `[STUB]` logger line

# 📊 Type of Change
- [x] Bug fix
- [x] New feature

# 📸 Screenshots / Recordings

# ⏱️ Estimated Review Time
- [ ] < 5 min
- [x] 5-15 min
- [ ] > 15 min


AysuKeskin (changes requested):
Thanks for the implementation. The overall structure looks clean, and separating the email-sending logic into `email_service.py` is a good design choice. The password reset flow also handles user enumeration and invalid token responses carefully.

However, I think there are a few issues that should be addressed before merging:

1. `reset_password` should be atomic. Currently, the password update and marking the token as used are two separate database writes. If something fails between them, the password may be changed while the token remains usable. This should be wrapped in `transaction.atomic()`.

2. The test settings should explicitly use Django’s in-memory email backend. Since the base settings now default to SMTP, `config/settings/test.py` should define `EMAIL_BACKEND = 'django.core.mail.backends.locmem.EmailBackend'`; otherwise, tests using `mail.outbox` may not behave correctly.

3. The new password reset endpoints do not seem to have rate limiting. Since existing auth views use `ScopedRateThrottle`, these endpoints should also have throttling to prevent reset-email spam, quota exhaustion, and token brute-force attempts.

4. Old unused password reset tokens are not invalidated when a new reset request is created. This means multiple valid reset links can exist for the same user. Before creating a new token, existing unused/unexpired tokens for that user should be deleted or marked as used.

5. In the latest commit, the `send_test_email` management command may print success even if sending fails, because the email service functions swallow exceptions. For a staging smoke test, the command should give real feedback when SMTP credentials or delivery fail.

6. `.env.example` mentions that `EMAIL_BACKEND` is overridden to console in `development.py`, but that change does not appear in the diff. This should either be implemented or the comment should be removed.

7. There are no tests for the new management command. Since this is new callable functionality, at least a basic test would be useful.

Minor notes:
- The `register_user` docstring is stale. It still says the verification code is not sent because email infrastructure is not implemented, but this PR now implements email sending.

- Some required docstrings are missing, especially for `PasswordResetToken`, `is_expired()`, `send_verification_email`, and `send_password_reset_email`.
- The deferred import in `validate_new_password` can be moved to the top of the file.
- Existing JWT refresh tokens are not invalidated after password reset. 

Overall, the feature is useful and mostly well-structured, but I think the atomicity, test email backend, rate limiting, token invalidation, and management command feedback issues should be fixed before merge.

AysuKeskin (changes requested):
Here is a issue I noticed while reviewing:


###  JWT refresh tokens are not invalidated after password reset


After a successful password reset, the user’s existing JWT refresh tokens remain valid. This means an old refresh token can still be used to obtain new access tokens even after the password has been changed.

This weakens the main security expectation of a password reset flow: if an account was compromised, resetting the password should also revoke existing authenticated sessions.

**Example attack scenario:**

1. An attacker gains access to a user’s account and obtains a refresh token.
2. The real user notices suspicious activity and resets their password.
3. The attacker’s refresh token remains valid.
4. The attacker can continue calling `POST /auth/token/refresh/` and receive new access tokens until the refresh token naturally expires.

**Suggested fix:**

Blacklist all outstanding refresh tokens for the user inside the existing `transaction.atomic()` block:

```python
from rest_framework_simplejwt.token_blacklist.models import (
    OutstandingToken,
    BlacklistedToken,
)

with transaction.atomic():
    token.user.set_password(new_password)
    token.user.save(update_fields=["password"])

    token.is_used = True
    token.save(update_fields=["is_used"])

    # Invalidate all outstanding refresh tokens for this user.
    for outstanding_token in OutstandingToken.objects.filter(user=token.user):
        BlacklistedToken.objects.get_or_create(token=outstanding_token)
```

If full session invalidation is deferred, I would at least add a TODO so this security gap is not missed later:

```python
# TODO: Blacklist all outstanding refresh tokens for token.user here.
# Without this, an attacker who obtained a refresh token before the reset
# can continue accessing the account until that token naturally expires.
```

---


Overall, the implementation looks good, but I think the refresh token invalidation issue should be addressed before merging because it has direct security implications.

KmalM:
Addressed the JWT refresh token invalidation issue from the latest review. All outstanding refresh tokens for the user are now blacklisted inside the `transaction.atomic()` block on password reset. Test added: `test_blacklists_outstanding_jwt_tokens`. All 365 tests pass.

AysuKeskin (approved):
Final state of the implementation looks solid to me. I do not see any visible issues at this point.

The previous concern I mentioned about JWT refresh tokens not being invalidated after a password reset has been resolved.

Approved.

PR #510, opened 2026-05-02
Feat/ support multiple tag filtering 

### LINKED ISSUE #506: [Enhancement][Backend] Support multiple tag filtering with AND logic
### Description

The current `/stories/feed/`, `/stories/map/`, and `/stories/search/` endpoints accept a single tag query parameter. The frontend now allows selecting multiple tag filters simultaneously. The backend needs to accept a tags parameter (comma-separated, e.g. tags=ottoman,galata) and apply AND logic — returning only stories that match all selected tags.  

### Acceptance Criteria

 - [ ] FeedQuerySerializer accepts tags as a comma-separated string (e.g. tags=ottoman,galata)                                                     
 - [ ] get_story_feed() and get_story_search() services apply a chained .filter() for each tag so AND logic is enforced
 - [ ] get_story_feed() in the map view also applies the same multi-tag filter                                                                     
 - [ ] Existing single-tag parameter continues to work                
 - [ ] Backend tests updated to cover multi-tag AND filtering                                                                                      
                                                            

### Meta (optional)

_No response_

AysuKeskin:
# 📝 Description 

Fixes #506 

This change was added based on an additional request from the frontend side, since the frontend needed to send multiple selected tags as repeated `tags` query parameters and receive results matching all selected tags.

This PR updates story filtering to support **multiple tag filters with AND logic** across the feed, map, and search endpoints.

Previously, story filtering supported a single `tag` query parameter. With this change, clients can pass repeated `tags` query parameters, such as:

`?tags=folklore&tags=ottoman-era`

The returned stories must contain **all** provided tags, not just one of them. This makes tag-based filtering more useful for discovery flows where users want to narrow results by combining multiple categories.

Main changes include:

- Replaced the single `tag` query parameter with a `tags` list parameter.
- Added AND-based tag filtering in `get_story_feed` and `get_story_search`.
- Updated feed, map, and search views to pass `tags` into the service layer.
- Updated serializer validation to reject empty tag lists.
- Added and updated tests for:
  - single tag filtering through `tags`
  - multiple tag filtering with AND semantics
  - case-insensitive tag matching
  - combination with year, location, geo, search, and map filters
  - duplicate prevention when stories have multiple tags

# 📊 Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Documentation

- [x] The project builds successfully.
- [x] Code follows the project style guidelines.
- [x] I have performed a self-review of my code.
- [x] Existing single-tag filtering behavior is preserved through the new `tags` parameter.
- [x] Feed endpoint supports repeated `tags` query parameters.
- [x] Map endpoint supports repeated `tags` query parameters.
- [x] Search endpoint supports repeated `tags` query parameters.
- [x] Multiple tag filters use AND logic, so returned stories must contain all selected tags.
- [x] Tag filtering remains case-insensitive.
- [x] Empty `tags` lists are rejected by serializer validation.
- [x] Filtering does not return duplicate stories when a story has multiple tags.
- [x] Tests were added/updated for serializers, services, and views.

# ⏱️ Estimated Review Time

- [ ] <5 min
- [x] 5–15 min
- [ ] >15 min

### DIFF
--- a/backend/apps/stories/serializers.py
+++ b/backend/apps/stories/serializers.py
@@ -322,8 +322,12 @@ class FeedQuerySerializer(serializers.Serializer):
     year_to = serializers.IntegerField(required=False)
     # Substring match against location_name — partial values like "galata" are valid
     location = serializers.CharField(required=False, allow_blank=False)
-    # Exact (case-insensitive) match against tag name — use the tag slug, e.g. "ottoman-era"
-    tag = serializers.CharField(required=False, allow_blank=False)
+    # AND filter — each tag name must match exactly (case-insensitive); all must be present on the story
+    tags = serializers.ListField(
+        child=serializers.CharField(allow_blank=False),
+        required=False,
+        allow_empty=False,
+    )
     latitude = serializers.FloatField(required=False, min_value=-90.0, max_value=90.0)
     longitude = serializers.FloatField(required=False, min_value=-180.0, max_value=180.0)
     # Minimum of 0.001 km — a zero radius would always return empty results.
@@ -406,7 +410,7 @@ class SearchQuerySerializer(FeedQuerySerializer):
     Validates query parameters for the story search endpoint.
 
     q is required. All filter params from FeedQuerySerializer (sort_by, year_from,
-    year_to, location, tag, latitude, longitude, radius_km) are inherited so search
+    year_to, location, tags, latitude, longitude, radius_km) are inherited so search
     results can be narrowed with the same filters as the feed.
     """
 
--- a/backend/apps/stories/services.py
+++ b/backend/apps/stories/services.py
@@ -61,7 +61,7 @@ def get_story_feed(
     year_from=None,
     year_to=None,
     location=None,
-    tag=None,
+    tags=None,
     latitude=None,
     longitude=None,
     radius_km=None,
@@ -102,8 +102,11 @@ def get_story_feed(
     if location:
         qs = qs.filter(location_name__icontains=location)
 
-    if tag:
-        qs = qs.filter(story_tags__tag__name__iexact=tag).distinct()
+    for tag in (tags or []):
+        # Each chained filter is a separate JOIN — AND semantics: story must have ALL tags
+        qs = qs.filter(story_tags__tag__name__iexact=tag)
+    if tags:
+        qs = qs.distinct()
 
     if latitude is not None and longitude is not None and radius_km is not None:
         qs = _apply_radius_filter(qs, latitude, longitude, radius_km)
@@ -215,7 +218,7 @@ def get_story_search(
     year_from=None,
     year_to=None,
     location=None,
-    tag=None,
+    tags=None,
     latitude=None,
     longitude=None,
     radius_km=None,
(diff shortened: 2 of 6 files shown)


KmalM (changes requested):
## Review

### Overview
Replaces the single `tag` query param with a `tags` list param supporting AND-logic multi-tag filtering across feed, map, and search. The chained-filter approach (`for tag in tags: qs = qs.filter(...)`) with a single `distinct()` after the loop is the correct Django pattern for AND semantics on M2M relationships. Tests are thorough.

### Blocking issue — `allow_empty=False` breaks all requests that omit `?tags=`

**File:** `backend/apps/stories/serializers.py` — `FeedQuerySerializer.tags`

In DRF 3.15.x, `ListField.get_value` has special handling for `QueryDict` (HTML input): when `tags` is **absent** from the query string, it returns `[]` (an empty list), **not** DRF's `empty` sentinel. Because `validate_empty_values` only short-circuits on the `empty` sentinel for `required=False` fields, the empty list is passed straight to `to_internal_value`, which raises `ValidationError` when `allow_empty=False`. 

This means **every request to the feed, map, and search endpoints that doesn't include any `?tags=` parameter returns 400** — i.e., the normal browsing flow is broken.

`allow_empty=False` is also redundant here: the only way for the list to be empty via query params is `?tags=` (an empty-string value), which is already rejected by `child=CharField(allow_blank=False)`.

**Fix:** remove `allow_empty=False` from the field declaration.

```python
tags = serializers.ListField(
    child=serializers.CharField(allow_blank=False),
    required=False,
)
```

Alternatively, add a custom `get_value` that returns `empty` when the key is absent from the `QueryDict`:

```python
class TagListField(serializers.ListField):
    def get_value(self, dictionary):
        if hasattr(dictionary, 'getlist'):
            values = dictionary.getlist(self.field_name)
            return values if values else serializers.empty
        return super().get_value(dictionary)
```

Otherwise the implementation and tests look clean. The DRF parsing concern raised above about `getlist()` is _not_ an issue — DRF already calls `getlist()` for `QueryDict` inputs, so `?tags=folklore&tags=ottoman-era` is parsed correctly.

AysuKeskin:
Thanks for the note. I checked this against DRF 3.15.2 and I don’t think the absent-query-param case returns `[]`.

For `QueryDict` input, `ListField.get_value()` does call `dictionary.getlist(self.field_name, [])`, but when the key is absent that gives an empty list and DRF then falls through to:

`html.parse_html_list(dictionary, prefix=self.field_name, default=empty)`

With no `tags[0]`, `tags[1]`, etc. keys present, `parse_html_list` returns the `empty` sentinel, not `[]`. Since the field has `required=False`, validation skips the field before `to_internal_value()` is reached, so `allow_empty=False` does not make ordinary requests without `?tags=` fail.

I also verified the behavior locally:

- no `tags` param: valid, field omitted
- `?tags=`: invalid because child `CharField(allow_blank=False)` rejects the blank value
- `?tags=folklore&tags=ottoman-era`: valid and parsed as `["folklore", "ottoman-era"]`

So I think `allow_empty=False` is not breaking feed/map/search requests when `tags` is absent.

KmalM:
You're right — I confirmed against the actual DRF 3.15.x source in the container:

```python
def get_value(self, dictionary):
    if self.field_name not in dictionary:
        if getattr(self.root, 'partial', False):
            return empty
    if html.is_html_input(dictionary):
        val = dictionary.getlist(self.field_name, [])
        if len(val) > 0:
            return val
        return html.parse_html_list(dictionary, prefix=self.field_name, default=empty)
    return dictionary.get(self.field_name, empty)
```

When `tags` is absent, `getlist` returns `[]`, the `len > 0` branch is skipped, and `parse_html_list` returns the `empty` sentinel. `required=False` then short-circuits before `to_internal_value` is reached, so `allow_empty=False` never fires. My concern was based on an older DRF implementation. The code is correct as-is.

KmalM (approved):
Verified the DRF 3.15.x behavior — the `allow_empty=False` concern in my earlier review was incorrect (see inline comment). The implementation is correct. Approving.

PR #512, opened 2026-05-02
fix: include tags in story feed API response

### LINKED ISSUE #470: [Bug][Backend] Include tags in story feed API response
### Description

 The frontend tag system is implemented and story cards are wired to display tag chips via `story.tags?.length > 0`. However, the `/stories/feed/` endpoint uses `StoryFeedSerializer` which does not currently include the tags field — so tag chips never render on feed story cards.            
  
  `StoryDetailSerializer` (used by `GET /stories/<pk>/`) already returns tags correctly through `StorySerializer`. The feed serializer needs the same field added. The queryset in StoryFeedView already calls `.prefetch_related('tags')`, so this is an additive change with no new database queries.

### Acceptance Criteria

  - [ ] `GET /stories/feed/`response includes a `tags` field on each story object
  - [ ] `tags` is a list of tag objects, each with `id`, `name`, `is_predefined`, and `story_count`                                                           
  - [ ] An empty list [] is returned when a story has no tags                                                                                       
  - [ ] `GET /stories/search/`response includes the same `tags` field (uses the same serializer)                                                       
  - [ ] No additional database queries are introduced (queryset already prefetches tags)                                                            
  - [ ] Existing feed endpoint tests are updated to cover the `tags` field in the response  

### Meta (optional)

_No response_

AysuKeskin:
# 📝 Description 
Fixes #470

This PR updates the story feed/search response so that each story card includes its associated tags.

This change was added based on a frontend-side need: the frontend requested tag data directly in the story card response in order to display selected/story-related tags without making an additional request.

Main changes include:

- Added `tags` to `StoryFeedSerializer` using `TagSerializer`.
- Included `tags` in the serialized story card fields.
- Added `prefetch_related('tags')` to feed and search querysets to avoid inefficient per-story tag queries.
- Added tests to verify:
  - feed response includes the `tags` field,
  - tagged stories return their tag data,
  - stories without tags return an empty `tags` list,
  - the expected feed card fields include `tags`.

# 📊 Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactoring
- [ ] Documentation

- [x] The project builds successfully.
- [x] Code follows the project style guidelines.
- [x] I have performed a self-review of my code.
- [x] Story feed response includes a `tags` field for each story card.
- [x] Story search response can also serialize tags through the shared feed serializer.
- [x] Tagged stories return their associated tag data.
- [x] Untagged stories return an empty `tags` list.
- [x] Tag serialization uses the existing `TagSerializer`.
- [x] Feed/search querysets prefetch tags to avoid N+1 queries.
- [x] Tests were added/updated for the feed response behavior.

# ⏱️ Estimated Review Time

- [x] <5 min
- [ ] 5–15 min
- [ ] >15 min

### DIFF
--- a/backend/apps/stories/serializers.py
+++ b/backend/apps/stories/serializers.py
@@ -236,6 +236,7 @@ class StoryFeedSerializer(serializers.ModelSerializer):
 
     user_has_liked = serializers.SerializerMethodField()
     user_has_saved = serializers.SerializerMethodField()
+    tags = TagSerializer(many=True, read_only=True)
 
     class Meta:
         model = Story
@@ -259,6 +260,7 @@ class StoryFeedSerializer(serializers.ModelSerializer):
             'user_has_liked',
             'user_has_saved',
             'submitted_at',
+            'tags',
         ]
         read_only_fields = fields
 
--- a/backend/apps/stories/services.py
+++ b/backend/apps/stories/services.py
@@ -85,7 +85,7 @@ def get_story_feed(
 
     sort_by — 'recent' (default) orders by submission date; 'popular' orders by like_count
     """
-    qs = Story.objects.filter(status=Story.STATUS_PUBLISHED)
+    qs = Story.objects.filter(status=Story.STATUS_PUBLISHED).prefetch_related('tags')
 
     if year_from is not None:
         qs = qs.filter(
@@ -235,6 +235,7 @@ def get_story_search(
     qs = (
         Story.objects
         .filter(status=Story.STATUS_PUBLISHED)
+        .prefetch_related('tags')
         .filter(Q(title__icontains=q) | Q(location_name__icontains=q))
     )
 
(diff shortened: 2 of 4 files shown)


KmalM (approved):
## Review

### Overview
Adds `tags` to `StoryFeedSerializer` so each story card includes its associated tags in feed and search responses. The `prefetch_related('tags')` additions in both service functions are the right call — without them, serializing N story cards would issue N+1 queries against the tags table.

### Implementation is clean

- `tags = TagSerializer(many=True, read_only=True)` correctly declares the field. DRF will use the prefetched `tags` queryset from `prefetch_related` during serialization, so no extra queries fire.
- The `Story.tags` M2M field (through `StoryTag`) is the correct accessor for `prefetch_related('tags')`. The existing filter path uses `story_tags__tag__...` (reverse FK), which is unrelated and unaffected.
- `prefetch_related` survives the full view pipeline — `.annotate()` (in `annotate_user_interactions`) and paginator slicing both preserve prefetch declarations.
- Tests cover tagged stories, untagged stories (empty list), and the expected card field set.

### Minor note

`TagSerializer` exposes `is_predefined` and `story_count` in addition to `id` and `name`. Both are low-sensitivity fields (`is_predefined` is a CMS flag, `story_count` is a denormalized counter), and `story_count` is already a model field so no extra query is needed. Reusing `TagSerializer` is a reasonable choice for now. If the feed payload size ever becomes a concern, a `TagSummarySerializer` with only `['id', 'name']` would be the solution.

No blocking issues.

PR #517, opened 2026-05-02
feat: email verification screen with 6-digit code

### LINKED ISSUE #179: [Feature][Frontend] Email verification screen with 6-digit code input
## Description

After a user registers, the backend now sends a 6-digit verification code to their email address (requirement 1.2.1.2). The frontend needs an intermediate verification screen that appears immediately after registration, before the user can access their account.

Currently, the registration flow goes directly from the registration form to the login page. This issue adds a verification step between registration and account activation.

### What needs to be built

1. **Verification screen:** A dedicated page/modal that appears after successful registration. It displays a message telling the user to check their email, and provides a 6-digit code input field (individual digit boxes for better UX, or a single text field with formatting).

2. **Code submission:** When the user enters the 6-digit code and submits, it calls the `POST /auth/verify-email` endpoint. On success, the account is activated and the user is redirected to the login page with a success toast ("Account verified! You can now log in.").

3. **Error handling:**
   - Invalid code: display inline error ("Invalid verification code. Please try again.")
   - Expired code: display error with a resend option ("Code has expired. Click to resend.")
   - Rate limiting: handle 429 responses gracefully

4. **Resend functionality:** A "Didn't receive the code? Resend" link/button that calls the resend endpoint. Include a cooldown timer (e.g., 60 seconds) to prevent spam clicking.

5. **API service:** Add `authService.verifyEmail(email, code)` and `authService.resendVerificationCode(email)` to the auth service layer.

### Why this matters

Email verification is a security requirement (1.2.1.2) that prevents fake account creation and ensures the user owns the email address they registered with. Without it, there's no way to reliably send password reset links or notifications.

## Acceptance Criteria

- [ ] Verification screen appears after successful registration
- [ ] 6-digit code input field with auto-focus and digit navigation
- [ ] Submit calls `POST /auth/verify-email` with email and code
- [ ] On success: redirect to login page with success toast
- [ ] Invalid code: inline error message displayed
- [ ] Expired code: error message with resend option
- [ ] Resend button calls resend endpoint with cooldown timer (60s)
- [ ] Loading state during verification submission
- [ ] API service functions: `authService.verifyEmail()`, `authService.resendVerificationCode()`
- [ ] Unit tests cover: code input rendering, submit flow, error states, resend cooldown, redirect on success

omarreis01:
# Description
Fixes #179

Inserts an email verification step between registration and account activation. After a successful register, the user is redirected to `/verify-email` where they enter the 6-digit code that the backend emails them. On success they are sent to `/login` with the original from-state preserved.

## What's added
- **`pages/VerifyEmailPage.jsx`** at `/verify-email` — 6 individual digit boxes with auto-focus, arrow-key navigation, backspace handling, paste support, inline error states, and a "Resend" button with a 60s cooldown timer.
- **`services/authService.js`** — `verifyEmail(email, code)` posting to `/auth/verify-email/`, `resendVerificationCode(email)` posting to `/auth/verify-email/resend/`.
- **Route** wired into `App.jsx`.
- **Unit tests** for the page (rendering, validation, success/error/expired/429, resend cooldown) and the new service methods.

## What's changed
- `RegisterPage.jsx` no longer attempts auto-login (which is incompatible with the new mandatory email-verified gate). It now redirects to `/verify-email` with `{ email, from }` in route state and shows an info toast prompting the user to check their email.
- `RegisterPage.test.jsx` updated to assert the new flow.

## Backend Context (informational, not a change request)
- Backend has the verification-code generation, email template, and `EmailVerificationCode` model (`apps/users/email_service.py`, `0007_passwordresettoken` migration). `RegisterView` already responds with `"Registration successful. Please verify your email."`.
- Backend does **not yet expose** `POST /auth/verify-email/` or `POST /auth/verify-email/resend/` — only `register/`, `login/`, `logout/`, `password-reset/`, and `password-reset/confirm/` are wired up in `apps/users/urls.py`.
- The frontend uses the endpoint names from issue #179. When backend wires them up under different paths the service URLs may need a one-line adjustment. Not a blocker for this PR — flagging for awareness.

# Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Documentation

# Screenshots / Recordings

# Estimated Review Time
- [ ] < 5 min
- [x] 5-15 min
- [ ] > 15 min

### DIFF
--- a/frontend/src/App.jsx
+++ b/frontend/src/App.jsx
@@ -7,6 +7,7 @@ import { ToastProvider } from "@/context/ToastContext";
 import { Toaster } from "@/components/ui/toaster";
 import LoginPage from "@/pages/LoginPage";
 import RegisterPage from "@/pages/RegisterPage";
+import VerifyEmailPage from "@/pages/VerifyEmailPage";
 import ProfileCompletionPage from "@/pages/ProfileCompletionPage";
 import NotFoundPage from "@/pages/NotFoundPage";
 import FeedPage from "@/pages/FeedPage";
@@ -27,6 +28,7 @@ function App() {
               <Route path="/map" element={<MapPage />} />
               <Route path="/login" element={<LoginPage />} />
               <Route path="/register" element={<RegisterPage />} />
+              <Route path="/verify-email" element={<VerifyEmailPage />} />
               <Route
                 path="/complete-profile"
                 element={(
--- a/frontend/src/pages/RegisterPage.jsx
+++ b/frontend/src/pages/RegisterPage.jsx
@@ -149,8 +149,11 @@ function RegisterPage() {
     try {
       await register(username, email, password, confirmPassword);
       toast.info("Account created! Check your email for a verification code.");
+      // Forward the password (held in route state, not in the URL) so
+      // VerifyEmailPage can auto-login the user the moment verification
+      // succeeds and send them straight to profile completion.
       navigate("/verify-email", {
-        state: { email, from: location.state?.from || null },
+        state: { email, password, from: location.state?.from || null },
         replace: true,
       });
     } catch (error) {
--- a/frontend/src/pages/RegisterPage.jsx
+++ b/frontend/src/pages/RegisterPage.jsx
@@ -14,7 +14,6 @@ import {
   CardFooter,
 } from "@/components/ui";
 import { register } from "@/services/authService";
-import { useAuth } from "@/hooks/useAuth";
 import { useToast } from "@/hooks/useToast";
 
 const INITIAL_FIELD_ERRORS = {
@@ -46,7 +45,6 @@ function validatePasswordStrength(password) {
 function RegisterPage() {
   const navigate = useNavigate();
   const location = useLocation();
-  const { login } = useAuth();
   const { toast } = useToast();
 
   const [username, setUsername] = useState("");
@@ -150,17 +148,11 @@ function RegisterPage() {
     setIsLoading(true);
     try {
       await register(username, email, password, confirmPassword);
-      try {
-        await login(email, password);
-        toast.success("Account created! Welcome!");
-        navigate("/complete-profile", {
-          state: { from: location.state?.from || null },
-          replace: true,
-        });
-      } catch {
-        toast.info("Account created! Please sign in to continue.");
-        navigate("/login", { state: { from: location.state?.from }, replace: true });
-      }
+      toast.info("Account created! Check your email for a verification code.");
+      navigate("/verify-email", {
+        state: { email, from: location.state?.from || null },
+        replace: true,
+      });
     } catch (error) {
       const msg = extractApiErrors(error);
       setApiError(msg);
--- a/frontend/src/pages/VerifyEmailPage.jsx
+++ b/frontend/src/pages/VerifyEmailPage.jsx
@@ -1,0 +1,311 @@
+import { useEffect, useRef, useState } from "react";
+import { useLocation, useNavigate, Link, Navigate } from "react-router-dom";
+import { Mail, Loader2, MapPin } from "lucide-react";
+
+import {
+  Button,
+  Label,
+  Card,
+  CardHeader,
+  CardTitle,
+  CardDescription,
+  CardContent,
+  CardFooter,
+} from "@/components/ui";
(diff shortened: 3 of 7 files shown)


sezindogan (commented):
## Overview                                                                                                                                      
                            
  Inserts an email-verification step between registration and account activation. After a successful register, the user is redirected to /verify-email, enters a 6-digit code, and is then sent to /login. Includes the new VerifyEmailPage, two new authService methods, routing changes, and a well-written test suite.
                                                                                                                                                                                                                                                                                        
### Blockers
                                                                                                                                                
  1. Branch name missing frontend/ prefix
                                                                                                                                                
  The branch is named feat/email-verification instead of feat/frontend-email-verification. Backend will likely need their own feat/email-verification branch when they implement the two missing endpoints (POST /auth/verify-email/ and POST /auth/verify-email/resend/).  
  The current name risks a naming collision and will cause confusion when both teams are working in the same repo. Recommend renaming to feat/frontend-email-verification before this is merged.

  2. Do not merge until backend endpoints are ready — this will break production registration                                                   
  
  The PR description already flags this honestly, but it bears emphasis as a hard blocker. RegisterPage has had its auto-login + redirect-to-profile flow removed. That means the moment this merges to main (and gets deployed), any user who registers will land on /verify-email and be stuck — the backend does not yet expose either verification endpoint. The deployed registration flow will be broken for all users until backend ships. This PR should either stay in draft or hold until the backend is implemented and wired.

  3. Post-verification destination is wrong — user is sent to /login instead of being auto-logged in                                            
  
  This is a UX and product correctness issue. After a successful verifyEmail() call the handler does:                                           
                            
  // VerifyEmailPage.jsx:219-220                                                                                                                
  toast.success("Account verified! You can now log in.");                                                                                       
  navigate("/login", { state: { from: fromAfterLogin }, replace: true });
                                                                                                                                                
  The user has just proven ownership of their email — forcing them to manually re-enter credentials on a separate page adds unnecessary friction and breaks the "newly registered user lands on profile completion" flow that existed before this PR.                                         
                                                                                                                                                
  The expected flow should be: verification success → auto-login with the credentials that were passed in route state → navigate to /complete-profile (or fromAfterLogin). This means:
  - The route state from RegisterPage should forward the password alongside the email (or the from destination), so VerifyEmailPage can call login(email, password) transparently on success.                                                                                              
  - On auto-login success, navigate to /complete-profile (same behaviour as the original happy path).
  - On auto-login failure (edge case), fall back to /login with from state, as the old code did.                                                
                                                                                                                                                
  Without this fix, the new registration funnel is strictly worse than what it replaces.                                                        
                                                                                                                                                                                                                                                                              
 ## Code Quality                                                                                                                                  
                            
 ### Positive
                                                                                                                                                
  - The digit-box UX (auto-advance, backspace, arrow-key navigation, paste distribution) is thorough and well-implemented.                      
  - extractApiError and isExpiredCodeError are cleanly separated and handle array vs. string responses gracefully.                              
  - The 60-second resend cooldown using setInterval with proper cleanup is correct.                                                             
  - aria-label, role="group", role="alert", aria-invalid, autoComplete="one-time-code" — solid accessibility coverage.                          
  - Test suite covers the golden path, error cases (400 invalid, expired, 429), resend cooldown, loading state, and the redirect guard. The     
  fake-timer test for the cooldown countdown is a nice touch.                                                                                   
                                                                                                                                                
  Minor issues                                                                                                                                  
                            
  - VerifyEmailPage.jsx:140 — the early-return guard (if (!email) return <Navigate .../>) runs after hooks are called, which is fine in React, but it sits below two useEffect calls. Reading order is slightly confusing; the guard could move earlier with an intermediate pattern (e.g., split into a shell component + inner component) to make the invariant obvious.                                                                
  - VerifyEmailPage.jsx:90 — EMPTY_DIGITS is a module-level constant but it is mutated indirectly via spread every time digits reset. This is safe as written, but naming it INITIAL_DIGITS would signal immutability more clearly.                                                         
  - authService.js — verifyEmail and resendVerificationCode use a raw axios.post without the authenticated client used elsewhere. Confirm this is intentional (unauthenticated user, so correct), but it is worth a comment since it is an exception to the pattern.                         
  - No authService.test.js coverage for network errors (no response, 500) on the two new methods. Low priority but keeps parity with the existing login/register tests.  They may be added when the backend is wired to the frontend

PR #521, opened 2026-05-02
feat: transactional email service, password reset, and email verification flow

### LINKED ISSUE #193: [Feature][Backend] Implement email verification flow for registration
## Description

The system currently supports registration with email and password, but the registration flow also needs email verification before account activation. According to the requirements, the system must validate the user’s email using a 6-digit verification code during registration. This issue covers the backend implementation of that verification flow.

The goal of this issue is to extend the registration process so that, on `POST /auth/register`, the system generates a 6-digit verification code, stores it securely, and sends it to the user’s email address. The newly created account should remain inactive or unverified until the code is confirmed through a dedicated verification endpoint.

A new endpoint, `POST /auth/verify-email`, should validate the submitted code and activate the account if the code is correct and has not expired. Verification codes should expire after 15 minutes. The implementation should also ensure that expired or already used codes cannot be reused.

This issue should include model/service updates if needed, code generation logic, expiry handling, serializer validation, API views, email trigger integration hook, and tests.

### Expected Behavior

**Registration step**
- Endpoint: `POST /auth/register`
- Creates the user in an inactive/unverified state
- Generates a 6-digit email verification code
- Stores the code with an expiry timestamp of 15 minutes
- Triggers sending the code to the user’s email

**Verification step**
- Endpoint: `POST /auth/verify-email`
- Accepts email and verification code
- Validates that the code matches the latest valid code for that user
- Rejects invalid, expired, or already used codes
- Marks the user as verified/active on success
- Prevents reuse of the same code after successful verification

### Notes
- This issue covers the verification flow logic itself; provider-level email delivery setup can be handled separately if needed.

## Acceptance Criteria

- [ ] `POST /auth/register` generates a 6-digit verification code
- [ ] Verification codes expire after 15 minutes
- [ ] Newly registered users are not activated until verification succeeds
- [ ] `POST /auth/verify-email` is implemented
- [ ] Invalid or expired codes are rejected
- [ ] Verified accounts are activated successfully
- [ ] Verification codes cannot be reused after successful verification
- [ ] Tests are added for successful and failing verification cases

## Meta (optional)

Deadline: 15-04-2026
Estimated Duration: 4 hours

### LINKED ISSUE #201: [Feature][Backend] Implement forgot-password and reset-password flow
## Description

The authentication system should support account recovery for users who forget their password. According to the project requirements, this should be handled through an email-based password reset flow in which the user requests a reset link and then submits a new password through a secure token-based process. This issue covers the backend implementation of that full flow.

The goal of this issue is to add a `POST /auth/forgot-password` endpoint that accepts an email address and triggers a password reset email when appropriate, along with a `POST /auth/reset-password` endpoint that validates a reset token and updates the user’s password securely. The `forgot-password` endpoint should be designed carefully so that it does not expose whether a given email address exists in the system, since returning different responses for existing and non-existing users could create a user enumeration risk.

When a valid reset request is made, the system should generate a reset token, associate it with the correct user, and send an email containing a password reset link. The reset token should be time-bound and single-use. The `reset-password` endpoint should validate the submitted token, reject expired or already used tokens, and update the password only if all validation checks pass. After a successful reset, the token must immediately become invalid so it cannot be reused.

This issue should include everything needed on the backend side to make the reset flow work reliably and securely, such as token generation and storage, validation logic, serializer and endpoint implementation, password update logic using the project’s secure hashing approach, and comprehensive tests. The implementation should also ensure that password rules applied during registration are enforced consistently during password reset so that users cannot bypass password requirements through the recovery flow.

### Expected Behavior

**Forgot password**
- Endpoint: `POST /auth/forgot-password`
- Accepts an email address
- Generates a password reset token for the matching user
- Sends a password reset link by email
- Returns a safe generic response regardless of whether the email exists

**Reset password**
- Endpoint: `POST /auth/reset-password`
- Accepts a reset token and a new password
- Validates token authenticity and usability
- Rejects invalid, expired, or already used tokens
- Updates the user password securely
- Invalidates the token immediately after successful use

## Acceptance Criteria

- [ ] `POST /auth/forgot-password` is implemented
- [ ] Password reset token is generated for valid accounts
- [ ] A reset link can be sent through the configured email flow
- [ ] `POST /auth/reset-password` is implemented
- [ ] Reset tokens are single-use
- [ ] Invalid or expired tokens are rejected
- [ ] Password is updated securely after valid reset
- [ ] Password validation rules are enforced on reset
- [ ] Tests are added for successful and failing reset flows

## Meta (optional)

Deadline: 15-04-2026
Estimated Duration: 4 hours

KmalM:
# 📝 Description
Fixes #193
Fixes #201

Implements the full transactional email infrastructure and both email-gated auth flows.

# 📊 Type of Change
- [x] New feature

## What's included

**Email service** (`apps/users/email_service.py`)
- Resend configured as the email provider via Django's `EMAIL_BACKEND`; `locmem` backend used in tests so `mail.outbox` assertions work without mocking
- `send_verification_email` and `send_password_reset_email` helpers; failures propagate to call sites which catch and log them so email outages don't block account creation

**Password reset flow** (closes #201)
- `POST /auth/password-reset/` — generates a single-use UUID token, invalidates any prior unused tokens, sends reset link; returns the same 200 response whether the email exists or not (anti-enumeration)
- `POST /auth/password-reset/confirm/` — validates token (expired and used both rejected with a unified error), updates password via `set_password`, blacklists all outstanding JWT tokens atomically
- Rate-limited with `ScopedRateThrottle`

**Email verification flow** (closes #193)
- `POST /auth/register` now creates users with `is_active=False`; a 6-digit code (expires in 15 min) is generated and emailed
- `POST /auth/verify-email/` — validates the code, marks it used, sets `is_active=True` and `is_email_verified=True` in a single atomic write; unified error for unknown email / wrong code / expired / reused
- Login blocked for unverified accounts via the existing `is_active` check in `login_user`

# 📸 Screenshots / Recordings
N/A — backend only

# ⏱️ Estimated Review Time
- [ ] < 5 min
- [x] 5-15 min
- [ ] > 15 min


AysuKeskin (changes requested):
The core flows look solid.
 

## Issues to Address

### 1. VerifyEmailView` has no rate throttle

`POST /auth/verify-email/` accepts a 6-digit verification code, but the endpoint currently has no `throttle_classes`.

Since the code has only `10^6` possible combinations, this creates a brute-force risk. `LoginView` and `PasswordResetRequestView` already use throttling, so this endpoint should follow the same pattern.

Suggested change:

```python
# views.py — VerifyEmailView
throttle_classes = [ScopedRateThrottle]
throttle_scope = "verify_email"
```

This also requires adding the corresponding throttle rate in settings.

---

### 2. No resend-verification endpoint

Verification codes expire after 15 minutes. If the user misses the email or the code expires, the account remains `is_active=False`, but there is currently no way to request a new verification code.

This creates a blocking UX issue because the user can become permanently unable to log in.

A `POST /auth/resend-verification/` endpoint would solve this. It should probably:

* Accept an email address
* Return a uniform response for anti-enumeration
* Generate a new verification code for inactive users
* Send the verification email
* Be rate-limited

---

### 3. `request_password_reset` invalidation and creation are not atomic

Current flow:

```python
PasswordResetToken.objects.filter(user=user, is_used=False).update(is_used=True)
token = PasswordResetToken.objects.create(user=user)
```

Under concurrent requests for the same user, two valid reset tokens could be created.

This can be fixed by wrapping the invalidation and creation in a transaction and locking the user row:

```python
with transaction.atomic():
    user = User.objects.select_for_update().get(pk=user.pk)
    PasswordResetToken.objects.filter(user=user, is_used=False).update(is_used=True)
    token = PasswordResetToken.objects.create(user=user)
```

This would better enforce the model invariant that only one valid reset token should exist at a time.

---

### 4. `_INVALID_VERIFICATION_CODE` is a shared exception singleton

Current pattern:

```python
_INVALID_VERIFICATION_CODE = ValidationError({
    "code": "Invalid or expired verification code."
})
```

Using a module-level exception instance is risky because exception objects may carry mutable state after being raised. A fresh exception should be created each time instead.

Suggested change:

```python
def _invalid_verification_code():
    return ValidationError({
        "code": "Invalid or expired verification code."
    })
```

Then raise it as:

```python
raise _invalid_verification_code()
```

---

### 5. `User.is_active` model default still allows active users by default

The model currently has:

```python
is_active = models.BooleanField(default=True)
```

`register_user` explicitly creates users with `is_active=False`, which is correct for the new verification flow. However, any direct `User.objects.create_user()` call will still create an active user by default.

This is especially easy to miss in future features or tests. The shared `conftest` fixture bypassing verification is fine, but it would be clearer and safer to make this explicit.

Consider changing the model default to:

```python
is_active = models.BooleanField(default=False)
```

Then update test fixtures that intentionally need active users to pass:

```python
is_active=True
```

---


## Summary

The main authentication flows are implemented correctly, and the PR significantly improves the security and completeness of the auth system.

Overall, this is a strong implementation. The remaining issues are mostly around closing edge cases and making the verification/reset flows safer under real-world usage.

KmalM:
All five review issues addressed:

1. **`VerifyEmailView` throttle** — added `throttle_classes = [ScopedRateThrottle]` and `throttle_scope = 'verify_email'` (10/min in prod, 10000/min in tests). `ResendVerificationView` shares the same scope since both endpoints are brute-force surfaces for the same 6-digit code.

2. **`POST /auth/resend-verification/`** — new endpoint that silently no-ops for unknown/already-verified addresses (anti-enumeration), generates a fresh `EmailVerificationCode`, and emails it. Rate-limited via `verify_email` scope.

3. **Atomic password reset** — `request_password_reset` now wraps invalidation + creation in `transaction.atomic()` with `select_for_update()` on the user row, preventing concurrent requests from each creating a valid token.

4. **`_INVALID_VERIFICATION_CODE` singleton** — converted to a factory function `_invalid_verification_code()` so each raise gets a fresh exception instance with no shared mutable state.

5. **`User.is_active` default** — changed to `default=False`. Migration `0008` added. All `create_user()` calls in tests and `conftest.py` fixtures updated to pass `is_active=True` explicitly. `test_models.py` updated to assert the new default.

AysuKeskin (changes requested):
Most of the previous issues have been addressed. The remaining concerns are below.

### 1. `UserManager.create_user` conflicts with the model default

`models.py:33`

```python
extra_fields.setdefault('is_active', True)
````

The model field now defaults to `False`, but `create_user()` resets `is_active` to `True` when it is not explicitly provided.

This means a direct call like this could still create an active but unverified user:

```python
User.objects.create_user(
    email="user@example.com",
    username="user",
    password="password123",
)
```

Passing `is_active=True` explicitly in `conftest.py` masks this issue in tests. The manager-level default should either be removed or changed to `False`.

---

### 2. `resend_verification` does not invalidate old codes


`services.py:95-96`

```python
code = EmailVerificationCode.generate_code()
EmailVerificationCode.objects.create(user=user, code=code)
```

If the user requests resend multiple times, old unused verification codes remain valid for 15 minutes.

The same pattern used for password reset tokens should be applied before creating the new code:

```python
EmailVerificationCode.objects.filter(
    user=user,
    is_used=False,
).update(is_used=True)
```

This would ensure that only the latest verification code remains usable.

---

### 3. Missing service-level unit test for `resend_verification`

There is no dedicated service-level unit test for `resend_verification`.

Since this is service-layer logic, it should be covered directly in `test_services.py`, not only through view tests. A dedicated `TestResendVerification` class would make this behavior easier to verify and maintain.

KmalM:
Addressed all three remaining issues:

1. **`UserManager.create_user` default** — Added `extra_fields.setdefault('is_active', False)` to `create_user` so the inactive-by-default contract is enforced at the manager level, not just via the model field. `create_superuser` already sets `is_active=True` before calling `create_user`, so its behavior is unchanged.

2. **`resend_verification` old code invalidation** — Added `EmailVerificationCode.objects.filter(user=user, is_used=False).update(is_used=True)` before creating the new code, matching the same pattern used by `request_password_reset`. Only the most recently issued code is now valid.

3. **Service-level tests for `resend_verification`** — Added `TestResendVerification` in `test_services.py` with 7 cases: creates new code, sends email, marks old codes as used, ensures only one valid code remains after repeated resends, and no-ops for unknown/active/already-verified addresses.

AysuKeskin (approved):
All issues I pointed out in the previous review have been addressed.

Summary of the current review:

| Issue                                  | Status                                                                  |
| -------------------------------------- | ----------------------------------------------------------------------- |
| VerifyEmailView throttle               | ✅ Uses `verify_email` scope with `10/min` limit                         |
| Resend endpoint                        | ✅ Anti-enumeration behavior added and endpoint is rate-limited          |
| Atomic password reset                  | ✅ Uses `transaction.atomic()` with `select_for_update()`                |
| Singleton exception                    | ✅ Replaced with a factory function                                      |
| `is_active` model default              | ✅ `default=False` and manager now uses `setdefault('is_active', False)` |
| Old verification codes not invalidated | ✅ `resend_verification` marks old codes with `update(is_used=True)`     |
| Missing resend service unit tests      | ✅ Covered by `TestResendVerification` with 7 test cases                 |

The implementation looks ready to merge. I do not see any remaining issues in terms of code quality or test coverage.

PR #531, opened 2026-05-04
Feat/frontend geographic location search

### LINKED ISSUE #273: [Feature][Frontend] Geographic location search via Nominatim geocoding
### Description

Currently the location filter does a text match against location_name. This means searching "Istanbul" returns nothing for a story located at "Hagia Sophia" even though it geographically falls within Istanbul.                                                                           
   
  Integrate the Nominatim geocoding API (OpenStreetMap, no key required) into the location filter field. When the user types a place name,  geocode it to a bounding box and pass lat_min, lat_max, lng_min, lng_max to the backend instead of the plain text location param. Fall back to the existing text search if Nominatim returns no result.

### Acceptance Criteria

- [ ] API is integrated successfully
- [ ]  Typing "Istanbul" in the location filter returns all stories whose coordinates fall within Istanbul's bounding box
- [ ]  Works for countries, cities, districts, and neighbourhoods                                                                                  
- [ ]  Falls back to text-based location_name search if Nominatim resolves nothing
- [ ]  Nominatim request is debounced (fires after the user stops typing, not on every keystroke)                                                  
- [ ]  Loading state shown while geocoding is in progress                                                                                          
- [ ]  storyService updated to accept and pass lat_min/max, lng_min/max params                                                                     
- [ ]  Unit tests for the geocoding logic and fallback behaviour 

### Meta (optional)

_No response_

sezindogan:
# 📝 Description
Fixes #273

Integrates Nominatim (OpenStreetMap) geocoding into the location filter field. Typing a place name now resolves to a bounding box and passes lat_min/lat_max/lng_min/lng_max to the backend instead of a plain text string, so searching "Istanbul" returns stories whose coordinates fall within Istanbul's bounds — not just stories with "Istanbul" in their location name. Falls back to text-based location_name search if Nominatim returns no result. This separates the purpose of the search bar and the filtering tab as intended, making each serve a difference purpose and expanding the capabilities of our search and filtering feature.                                                  

 ### New features

  - geocodingService.js: geocodeLocation (single bbox result) and searchLocationSuggestions (up to 5 results with title, subtitle, bbox)               
  - useLocationSuggestions hook: debounced autocomplete (400 ms, min 3 chars) with clearSuggestions
  - useGeocoding hook: debounced single-result geocoding (600 ms) kept as a standalone utility                                                         
  - Autocomplete dropdown in FilterPanel: suggestions appear as the user types, clicking one locks in the bbox and fills the input; dropdown never opens on mount or after selection                                                                                                                    
  - storyService: getStories and getMapStories now accept latMin/latMax/lngMin/lngMax and pass bbox params to the backend, falling back to text location when no bbox is present                                                                                                                     
  - useFilterState: reads/writes lat_min/lat_max/lng_min/lng_max URL params alongside existing filter params
  - SearchFilter: removing the location chip also clears bbox URL params atomically                                                                    
                                                                                                                                                       
###  Modified files                                                                                                                                       
                                                                                                                                                       
  - src/services/storyService.js — bbox params in getStories and getMapStories                                                                         
  - src/hooks/useFilterState.js — bbox URL param support
  - src/components/SearchFilter/FilterPanel.jsx — autocomplete dropdown, lockedBbox/suggestionsQuery state                                             
  - src/components/SearchFilter/SearchFilter.jsx — forward bbox through handleFilterApply, clear bbox on location chip removal                         
  - src/pages/FeedPage.jsx — destructure and pass bbox from filter state                                                                               
  - src/pages/MapPage.jsx — same                                                                                                                       
                                                                                                                                                       
 ### Tests                                                                                                                                                
                                                            
  - geocodingService.test.js — geocodeLocation and searchLocationSuggestions (empty/short queries, valid result, missing bbox, bad coordinates, HTTP   
  errors)
  - useGeocoding.test.js — debounce timing, null/error fallback, rapid-typing cancellation                                                             
  - useLocationSuggestions.test.js — debounce, min-chars guard, error fallback, clearSuggestions                                                       
  - storyService.test.js — bbox params passed instead of text location, text fallback when no bbox                                                     
  - FilterPanel.test.jsx — updated mock (useLocationSuggestions) and assertions to include bbox fields 

# 📊 Type of Change
- [ ]  Bug fix
- [x]  New feature
- [ ]  Refactoring
- [ ]  Documentation
 
# 📸 Screenshots / Recordings
<!-- If applicable !-->
<img width="1440" height="900" alt="Screenshot 2026-05-04 at 15 05 55" src="https://github.com/user-attachments/assets/38f99915-b5ce-416f-b141-8fcbbbb0de61" />

# ⏱️ Estimated Review Time
- [ ] < 5 min
- [ ] 5-15 min
- [x] > 15 min


omarreis01 (commented):
## PR Review: feat/frontend geographic location search

### Summary
Solid feature. Replaces text-only location filtering with Nominatim geocoding so "Istanbul" actually filters by Istanbul's bounding box rather than substring-matching the location_name string. Code is clean, the new hooks/services are well isolated, and test coverage for the new units is genuinely good.

The frontend↔backend contract checks out — backend already supports `lat_min/lat_max/lng_min/lng_max` on `/stories/feed/`, `/stories/map/`, and `/stories/search/`.

A few real issues are worth addressing before merge — none catastrophic, but two of them silently corrupt user state.

---

### Frontend

**🚫 Blocker — bbox is silently dropped when re-opening the filter panel**

`SearchFilter.jsx` keys `FilterPanel` on `\`${yearFrom}-${yearTo}-${location}\``, so the panel unmounts on every apply. On remount, `lockedBbox = null` and `suggestionsQuery = ""`. If the user filters by "Istanbul" (with bbox), then later opens the panel just to change the year and clicks Apply without re-typing the location, `effectiveBbox = lockedBbox ?? suggestions[0]?.bbox ?? null` evaluates to `null`, and `handleFilterApply` writes `lat_min: ""`, `lat_max: ""`, etc., which `useFilterState.setFilters` deletes from the URL.

Net effect: changing year silently downgrades the filter from bbox-based to substring-based, without the user touching the location field.

Fix options: pass the current `latMin/latMax/lngMin/lngMax` from `useFilterState` into `FilterPanel` as a prop and initialize `lockedBbox` from it; OR preserve the existing bbox unless the user actually edits the location input.

**⚠️ Race condition in `useGeocoding` and `useLocationSuggestions`**

The cleanup only `clearTimeout`s the debounce timer. If the timer has already fired and the `await fetch` is in flight when the query changes, the in-flight request is not cancelled and there's no `cancelled` flag — so a slow earlier response can resolve after a fast later one and overwrite state with stale data. Standard fix:

```js
useEffect(() => {
  let cancelled = false;
  const timer = setTimeout(async () => {
    try {
      const result = await geocodeLocation(query);
      if (!cancelled) setBbox(result);
    } catch {
      if (!cancelled) setBbox(null);
    } finally {
      if (!cancelled) setIsLoading(false);
    }
  }, DEBOUNCE_MS);
  return () => { cancelled = true; clearTimeout(timer); };
}, [query]);
```

Visible mostly on slow connections / fast typers, but real.

**⚠️ Nominatim usage from the browser**

Two concerns with hitting `nominatim.openstreetmap.org` directly from the client:
1. The `User-Agent` header set in `NOMINATIM_HEADERS` is a [forbidden header name](https://developer.mozilla.org/en-US/docs/Glossary/Forbidden_header_name) and is silently dropped by all browsers — Nominatim won't actually see `LocalHistoryStoryMap/1.0`. Their usage policy explicitly requires a meaningful User-Agent identifying the application, which is impossible from `fetch`. Worth either removing the header (it's misleading) or proxying through the backend so a real UA can be set.
2. Their policy caps usage at ~1 req/sec/IP. With a 400 ms suggestions debounce, fast typers will exceed this. Not a blocker for an MVP, but expect occasional 429s in heavy demos.

Not blocking the PR, but flagging — proxying through the Django backend (and caching responses) is the standard fix for both issues if this feature stays in production.

**Non-blocking notes:**
- Suggestions dropdown is mouse-only: no keyboard navigation (ArrowUp/Down/Enter/Escape), no `aria-activedescendant`, no click-outside dismissal. The `role`/`aria-expanded` attributes are correctly set, but a screen-reader user can't actually pick a suggestion.
- If the user types but doesn't click a suggestion and hits Apply, `effectiveBbox = lockedBbox ?? suggestions[0]?.bbox ?? null` quietly applies the *first* suggestion's bbox. If suggestions are still loading, it falls back to text. This means clicking Apply quickly vs. waiting 400 ms can give meaningfully different results from the same typed input — surprising UX. Consider either always requiring an explicit pick, or showing the auto-selected match in the input before Apply.
- `useFilterState.js`: bbox URL param parsing has no direct unit test (added everywhere else, just not here). Not blocking, but a small `it("parses bbox query params")` would round it out.

**Tests:** `geocodingService.test.js`, `useGeocoding.test.js`, `useLocationSuggestions.test.js`, `storyService.test.js`, and the `FilterPanel.test.jsx` mock update are all well-structured — debounce, error fallback, empty/short queries, invalid coordinates, race-cancellation timing, bbox vs. text fallback are all covered. Nice work here.

---

I read `backend/apps/stories/views.py`, `serializers.py`, and `urls.py` to verify the API contract. Everything the frontend needs is already in place — no backend coordination required for this PR.

| Check | Status |
|-------|--------|
| `/stories/feed/`, `/stories/map/`, `/stories/search/` URL paths | ✅ exist |
| `lat_min/lat_max/lng_min/lng_max` accepted on all three endpoints | ✅ via `FeedQuerySerializer` (lines 340–343) and `SearchQuerySerializer` (inherits) |
| Bounds validation (-90..90 lat, -180..180 lng) | ✅ |
| Cross-validation (`lat_min ≤ lat_max`, `lng_min ≤ lng_max`) | ✅ (serializer `validate()`) |
| Bbox actually filters the queryset | ✅ `apply_bbox_filters(qs, params)` in `views.py` |
| Text `location` fallback still supported when no bbox sent | ✅ both pathways live in parallel |
| Auth | ✅ all three endpoints `AllowAny`, frontend doesn't need a token |

---

- `GET /stories/feed/` — ✅ bbox params (`lat_min`, `lat_max`, `lng_min`, `lng_max`) and text `location` both supported; frontend's mutually-exclusive logic matches
- `GET /stories/search/` — ✅ same as above; `q` required (frontend gates on `q?.trim()`)
- `GET /stories/map/` — ✅ same as above; FeatureCollection response shape consumed correctly

---

### Verdict
**REQUEST CHANGES**

Blockers (please fix before merge):
- FilterPanel silently drops bbox when re-opened to edit only the year — initialise `lockedBbox` from the current URL state via a prop from `SearchFilter`.

Should-fix (strongly recommended, but I'll defer to you on whether to bundle here or follow-up):
- Add a `cancelled` flag in `useGeocoding` / `useLocationSuggestions` to prevent stale-response overwrites.
- Either remove the misleading `User-Agent` header (browsers drop it) or proxy Nominatim through the backend so the policy is actually honoured.

Nice-to-haves:
- Keyboard navigation + click-outside dismiss for the suggestions dropdown.
- Decide whether silent auto-pick of `suggestions[0]` on Apply is intended UX.
- Direct unit test for `useFilterState`'s bbox parsing.

Genuinely good PR overall — clean separation of concerns, real test coverage, and the contract with the backend is exactly right. ✅

omarreis01 (changes requested):
Frontend blocker noted in the comment above: FilterPanel silently drops the bbox URL params when the panel is re-opened to edit only the year. Two should-fix items (race condition in geocoding hooks, Nominatim browser usage) detailed there as well. Backend contract is fine.

sezindogan:
Addressed all review feedback:                                                                                                                       
                                                                                                                                                       
 ### Blocker fixed — bbox silently dropped on panel reopen                                                                                                
  SearchFilter now reads latMin/latMax/lngMin/lngMax from useFilterState and passes them down to FilterPanel as props. FilterPanel initialises lockedBbox from those props on mount, so reopening the panel to change only the year preserves the previously selected bbox. The key was also updated to include bbox values so external clears (e.g. resetting all filters) still remount the panel correctly.                                           
                                                                                                                                                       
 ### Race condition fixed                                                
  Added a cancelled flag in both useGeocoding and useLocationSuggestions. In-flight fetch responses that arrive after the query has changed are now discarded instead of overwriting state with stale data.                                                                                              
   
 ### Misleading User-Agent header removed                                                                                                                 
  Removed from geocodingService.js — it's a forbidden header name and was being silently dropped by the browser anyway. A backend proxy issue has been filed to properly address Nominatim's identification and rate-limit requirements (see #534 ). After this issue is implemented by backend, all Nominatim API calls will be updated accordingly.                                                    
                                                                      
 ### Silent auto-pick removed                                                                                                                             
  effectiveBbox in handleApply is now just lockedBbox — the suggestions[0]?.bbox fallback is gone. Typing a location without explicitly selecting a suggestion always falls back to text filtering, consistently regardless of whether suggestions have loaded.                                          
                                                                      
 ### Keyboard navigation + click-outside added                                                                                                            
  The suggestions dropdown now supports ArrowUp/Down to navigate, Enter to select, Escape to dismiss. Each suggestion has an id and aria-selected; the input has aria-activedescendant wired up. Clicking outside the location field dismisses the dropdown.                                                
   
  Performance improvements                                                                                                                             
  - Debounce reduced from 400ms to 250ms                              
  - Module-level cache added to geocodingService.js — repeated or backspaced-then-retyped queries return instantly without hitting Nominatim           
   
  Tests                                                                                                                                                
  - Added bbox preservation test to FilterPanel.test.jsx              
  - Added bbox URL param parsing tests to useFilterState.test.jsx                                                                                      
  - Updated geocodingService.test.js to call clearCache() in beforeEach (new export)
  - Updated useLocationSuggestions.test.js timer values to match new 250ms debounce                                                                    
  - All 665 tests pass, lint clean

omarreis01 (approved):
## Re-review after revisions ✅

All previously-flagged issues are addressed, and you went beyond the original scope on the Nominatim concern. Verified each change against the latest diff and against `main`:

### Blockers — all resolved
- 🚫→✅ **Bbox preserved on panel reopen.** `SearchFilter.jsx` now reads `latMin/latMax/lngMin/lngMax` from `useFilterState` and forwards them as props; `FilterPanel.jsx` initialises `lockedBbox` from those props on mount. The remount key was also extended to include the bbox values, so external clears still reset correctly. New test `preserves bbox props in onApply when user only changes year without re-typing location` directly covers this regression. Confirmed.

### Should-fix — all resolved
- ⚠️→✅ **Race condition.** `useGeocoding.js` and `useLocationSuggestions.js` now use a `cancelled` flag inside the effect and check it before every `setState`. Stale in-flight responses are dropped instead of overwriting fresh state. Clean implementation.
- ⚠️→✅ **Nominatim browser usage.** Better than the requested fix — instead of just removing the misleading `User-Agent` header, you proxied through the Django backend (`/geocode/` and `/geocode/suggestions/`). I cross-checked `origin/main`: `apps/geocoding/{urls,views,services}.py` are live, response shapes (`{lat_min, lat_max, lng_min, lng_max}` for single, `{id, title, subtitle, bbox}` array for suggestions) match the frontend's parsing in `geocodingService.js` exactly. Server-side caching + UA + rate-limit handling now live in one place. Issue #534 is closed.

### Nice-to-haves — all done
- ✅ Keyboard nav (ArrowUp/Down/Enter/Escape) + `aria-activedescendant` + click-outside dismiss in `FilterPanel.jsx`.
- ✅ Silent `suggestions[0]` auto-pick removed — `effectiveBbox = lockedBbox` only. Apply now behaves consistently regardless of suggestion-load timing.
- ✅ Direct `useFilterState` bbox-parsing tests added (parses numbers, returns null when absent).
- ✅ Module-level cache in `geocodingService.js` with `clearCache()` export for tests.
- ✅ Suggestions debounce 400 → 250 ms (single-geocode stays at 600 ms).

### API Contract (Frontend ↔ Backend) — re-verified
- `GET /geocode/?q=…` — ✅ returns `{lat_min, lat_max, lng_min, lng_max}` or `null`; frontend remaps to camelCase correctly.
- `GET /geocode/suggestions/?q=…` — ✅ returns up to 5 `{id, title, subtitle, bbox}`; backend already enforces 3-char minimum and lower-cases the query, frontend gates on the same minimum client-side.
- `GET /stories/{feed,search,map}/` — ✅ unchanged from original review; bbox params accepted, mutually-exclusive logic with text `location` still correct.

### Tiny non-blocking nit
`FeedPage.jsx` line 16 — destructuring order is `latMin, latMax, lngMin, tags, lngMax, …`, which still works (object destructuring is name-based) but reads a bit oddly. Cosmetic only, not worth a follow-up.

---

**Verdict: APPROVE.** Excellent turnaround — every blocker addressed, plus the Nominatim proxy is a meaningful upgrade over the minimum fix. Ship it. 🚀

PR #533, opened 2026-05-04
fix: unify password validation across registration and reset

### LINKED ISSUE #532: [Bug][Backend] Password reset rejects passwords that registration accepts
### Overview

There is a discrepancy between the password validation logic used during User Registration versus Password Reset. The registration process uses a custom, less restrictive validator, while the password reset process utilizes Django’s full suite of built-in validators. This inconsistency allows users to create accounts with "common" passwords that they are later forbidden from using during a reset, leading to a confusing user experience.

### Discrepancy Details

- RegisterSerializer: Uses validate_password_strength(). This only checks for basic complexity (length, digits, uppercase, and lowercase) and bypasses Django's internal security checks.

- PasswordResetConfirmSerializer: Uses Django's built-in validate_password(). This runs all AUTH_PASSWORD_VALIDATORS configured in settings, including the CommonPasswordValidator.

### Steps to Reproduce

- Register a new account using a password that meets complexity requirements but is technically common (e.g., 123456Aa).

- Observe that registration is successful.

- Initiate a password reset flow for that same account.

- Attempt to set the new password to another common password (123456Ss).

### Observed Behavior

- The Password Reset fails with a validation error: "This password is too common."

- The user is blocked from using credentials that the system previously accepted, creating an "unexpected" barrier with no explanation as to why the initial registration was allowed.

### Expected Behavior

- Validation logic should be centralized and consistent across the entire application.

- RegisterSerializer should be updated to use Django’s built-in validate_password() to ensure CommonPasswordValidator is applied at the point of account creation.

- The user should encounter the same security constraints regardless of the entry point (Registration vs. Reset).

I suggest using our custom designed function as it was designed with our application constraints in mind. Choosing this approach will also minimize the work needed by avoiding frontend and mobile issues with the updated backend. 

### Acceptance Criteria

  - [ ] Both registration and password reset enforce the same password rules
  - [ ] Either: RegisterSerializer is updated to also call validate_password(), or PasswordResetConfirmSerializer is switched to validate_password_strength() — whichever the team decides is the correct baseline
  - [ ] If CommonPasswordValidator is kept, the frontend password strength checklist on ResetPasswordPage will need to be updated to surface the "too common" error to the user. An issue for this should be opened to let the frontend and mobile temas know if this is the preferred solution approach.
  - [ ] Existing backend tests pass; new test added confirming both flows reject/accept the same inputs

### Meta (optional)

Effort: S
Deadline: no-deadline
Priority: P2

KmalM:
# 📝 Description
Fixes #532

Both registration and password reset now enforce identical password rules by using the same `validate_password_strength()` validator. Previously, `PasswordResetConfirmSerializer` called Django's built-in `validate_password()`, which applies `CommonPasswordValidator` (among others) — rejecting passwords like `123456Aa` that registration happily accepted.

The fix: switch `PasswordResetConfirmSerializer.validate_new_password` to use the custom `validate_password_strength()`, which checks only the four criteria from Req. 1.2.1.4 (length ≥ 8, digit, uppercase, lowercase).

# 📊 Type of Change
- [x] Bug fix

# 📸 Screenshots / Recordings

# ⏱️ Estimated Review Time
- [x] < 5 min

### DIFF
--- a/backend/apps/users/serializers.py
+++ b/backend/apps/users/serializers.py
@@ -1,5 +1,3 @@
-from django.contrib.auth.password_validation import validate_password
-
 from rest_framework import serializers
 
 from apps.users.models import Follow, User
@@ -298,7 +296,7 @@ class PasswordResetConfirmSerializer(serializers.Serializer):
     new_password_confirmation = serializers.CharField(write_only=True)
 
     def validate_new_password(self, value):
-        validate_password(value)
+        validate_password_strength(value)
         return value
 
     def validate(self, attrs):
--- a/backend/apps/users/tests/test_views.py
+++ b/backend/apps/users/tests/test_views.py
@@ -1147,6 +1147,67 @@ class TestPasswordResetConfirmView:
         })
         assert response.status_code == status.HTTP_400_BAD_REQUEST
 
+    def test_common_but_complex_password_is_accepted(self, client, user):
+        # Regression for #532: password reset must use the same validator as
+        # registration — CommonPasswordValidator must NOT be applied here.
+        token = self._make_token(user)
+        response = client.post(self.url, {
+            'token': str(token.token),
+            'new_password': '123456Aa',
+            'new_password_confirmation': '123456Aa',
+        })
+        assert response.status_code == status.HTTP_200_OK
+
+
+# ── Password validation consistency (issue #532) ─────────────────────────────
+
+@pytest.mark.django_db
+class TestPasswordValidationConsistency:
+    """Both registration and password reset must enforce identical password rules."""
+
+    REGISTER_URL = '/auth/register/'
+    RESET_URL = '/auth/password-reset/confirm/'
+
+    def _make_token(self, user):
+        from apps.users.models import PasswordResetToken
+        return PasswordResetToken.objects.create(user=user)
+
+    def test_common_password_accepted_by_both_flows(self, client, user):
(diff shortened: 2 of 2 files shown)


AysuKeskin (approved):
This PR fixes a UX-breaking inconsistency:
 PasswordResetConfirmSerializer was calling Django's  built-in validate_password(), which runs CommonPasswordValidator and rejects passwords that RegisterSerializer — using the custom validate_password_strength() — happily accepts. The fix replaces the built-in call with the   custom validator and adds regression tests.

  ---
  Code Quality

  backend/apps/users/serializers.py

  - The one-line change is correct and minimal. validate_password_strength was already imported at line 4 via    
  common.validators, so no new import is needed. The stale from django.contrib.auth.password_validation import 
  validate_password import is also cleaned up. No issues here.                                                   
                  
  backend/common/validators.py

  - The validator is well-structured, has a docstring, and raises serializers.ValidationError with a list of     
  messages, matching how DRF error responses are normalized across the project. No changes needed.
                                                                                                                 
  ---             
  Test Coverage
               
  - test_common_but_complex_password_is_accepted is a good, targeted regression test placed correctly inside the
  existing TestPasswordResetConfirmView class.                                                                   
  - TestPasswordValidationConsistency adds cross-flow coverage asserting both registration and reset
  accept/reject identical inputs — exactly what the issue asks for.                                              
  - Both negative and positive cases are present.

---

The fix is correct, minimal, and follows project conventions.

PR #537, opened 2026-05-06
Feat/admin ban user and delete tag endpoints

### LINKED ISSUE #231: [Feature][Backend] Implement admin user ban and tag deletion endpoints
### Description

Admins need endpoints to disable user accounts and remove tags as part of the moderation toolkit. Banning a user should disable their ability to log in by setting `is_active` to false, without deleting their account or content. Removing a tag should dissociate it from all stories via the `StoryTag` through-table before deletion to preserve referential integrity and decrement affected story tag counts.

**Scope:**

* implement `PATCH /admin/users/{id}/ban/` endpoint (admin only)
  * set user `is_active` to false to prevent login
  * preserve the user record and their content
* implement `DELETE /admin/tags/{id}/` endpoint (admin only)
  * remove all `StoryTag` associations for the tag before deleting
  * decrement `Tag.story_count` and affected story tag counts accordingly
* add serializers for ban response and tag deletion response
* add service-layer logic for user ban and tag removal
* return 403 for non-admin requests
* return 404 when the target user or tag does not exist
* register endpoints under the admin route
* add serializer, service, and endpoint tests

**Suggested branch name:**
`feat/admin-user-tag-management`

**Depends on:**
tag system (issue: #209)

### Acceptance Criteria

* [ ] `PATCH /admin/users/{id}/ban/` sets `is_active` to false for the target user
* [ ] Banned users cannot log in; their account and content are preserved
* [ ] `DELETE /admin/tags/{id}/` removes all `StoryTag` associations before deleting the tag
* [ ] Non-admin requests receive a 403 response
* [ ] 404 is returned when the target user or tag does not exist
* [ ] URLs are wired under the admin route
* [ ] Serializer, service, and endpoint tests are added and passing

### Meta (optional)

Deadline: 29-04-2026
Estimated Time: 3 hours

AysuKeskin:
# 📝 Description

Fixes #231

This PR adds new admin moderation capabilities for banning users and deleting tags.

The purpose of this change is to give admins safer and more direct tools for moderation. Admins can now disable problematic user accounts without deleting their data, and they can remove tags from the system together with their story associations when needed.

This PR adds an admin-only user ban endpoint:

`PATCH /moderation/users/{id}/ban/`

This endpoint disables the target user by setting `is_active=False`. The user account itself is preserved, and the user's existing content is not deleted. This makes the ban operation safer because it prevents future access while keeping the related data available for moderation, auditing, or future review.

Implemented user moderation changes:
- Added `AdminUserBanView` for admin-only user banning.
- Added the `ban_user` service function to keep the ban logic outside the view.
- Made `ban_user` idempotent, so banning an already banned user is treated as a no-op and still returns the user safely.
- Added `UserBanResponseSerializer` to return the banned user's basic information.
- Preserved the user account and existing stories after banning.

This PR also adds an admin-only tag deletion endpoint:

`DELETE /moderation/tags/{id}/`

This endpoint allows admins to delete a tag through the moderation route. When a tag is deleted, its related `StoryTag` associations are explicitly removed before deleting the tag itself. This ensures the cleanup happens in a controlled way and allows related delete signals to run properly.

Implemented tag moderation changes:
- Added `AdminTagDeleteView` for admin-only tag deletion.
- Updated the `delete_tag` service to explicitly delete related `StoryTag` rows before deleting the tag.
- Wrapped tag deletion logic in `transaction.atomic()` so the operation is handled safely.
- Registered the new moderation routes in `backend/config/urls.py`.

This PR also adds service and view tests covering:
- Successful admin user banning.
- Idempotent behavior when banning an already banned user.
- Preservation of the user account and stories after banning.
- Permission checks for admin, registered user, and guest users.
- `404` behavior for nonexistent users and tags.
- Successful admin tag deletion.
- Removal of story-tag associations.
- Tag deletion with no associations and multiple associations.

# 📊 Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Documentation
 
- [x] Code follows project style guidelines
- [x] Code is commented where necessary
- [x] Self-review performed
- [x] Tests added/updated
- [x] Tests pass locally

# ⏱️ Estimated Review Time

- [ ] < 5 min
- [x] 5–15 min
- [ ] > 15 min

### DIFF
--- a/backend/apps/tags/admin_urls.py
+++ b/backend/apps/tags/admin_urls.py
@@ -1,0 +1,7 @@
+from django.urls import path
+
+from apps.tags.admin_views import AdminTagDeleteView
+
+urlpatterns = [
+    path('tags/<int:pk>/', AdminTagDeleteView.as_view(), name='admin-tag-delete'),
+]
--- a/backend/apps/tags/admin_views.py
+++ b/backend/apps/tags/admin_views.py
@@ -1,0 +1,20 @@
+from django.shortcuts import get_object_or_404
+from rest_framework import status
+from rest_framework.response import Response
+from rest_framework.views import APIView
+
+from common.permissions import IsAdminUser
+
+from .models import Tag
+from .services import delete_tag
+
+
+class AdminTagDeleteView(APIView):
+    """DELETE /moderation/tags/{id}/ — remove a tag and all story associations (admin only)."""
+
+    permission_classes = [IsAdminUser]
+
+    def delete(self, request, pk):
+        tag = get_object_or_404(Tag, pk=pk)
+        delete_tag(tag)
+        return Response(status=status.HTTP_204_NO_CONTENT)
--- a/backend/apps/tags/services.py
+++ b/backend/apps/tags/services.py
@@ -45,8 +45,15 @@ def create_tag(validated_data: dict, *, is_predefined: bool = False) -> tuple:
 
 
 def delete_tag(tag: Tag) -> None:
-    """Delete the tag. StoryTag rows cascade via FK."""
-    tag.delete()
+    """
+    Delete a tag and all its story associations.
+
+    StoryTag rows are deleted explicitly so that post_delete signals fire and
+    Tag.story_count is decremented before the tag itself is removed.
+    """
+    with transaction.atomic():
+        StoryTag.objects.filter(tag=tag).delete()
+        tag.delete()
 
 
 def attach_tags_to_story(story, tag_ids: list) -> None:
--- a/backend/apps/tags/tests/test_services.py
+++ b/backend/apps/tags/tests/test_services.py
@@ -122,28 +122,51 @@ class TestCreateTag:
 
 @pytest.mark.django_db
 class TestDeleteTag:
-    def test_deletes_tag(self):
-        tag = Tag.objects.create(name='svc-to-delete')
-        delete_tag(tag)
-        assert not Tag.objects.filter(name='svc-to-delete').exists()
-
-    def test_cascade_deletes_story_tags(self):
+    def _make_story(self, email, username):
         from apps.stories.models import Story
         from apps.users.models import User
-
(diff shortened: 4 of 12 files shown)


KmalM (changes requested):
## Review

The implementation is clean and well-structured. One unintentional regression needs to be restored before this can be merged.

### Blocking

**Accidentally removed test** — `test_missing_password_returns_400` was deleted from `TestDeleteAccountView` in `apps/users/tests/test_views.py`. It tests that `DELETE /users/me/` with an empty body returns 400 — unrelated to the ban feature, so this looks like a copy/paste casualty when the new `AdminUserBanView` section was inserted. Please restore it:

```python
def test_missing_password_returns_400(self, auth_client):
    response = auth_client.delete(self.url, {}, format='json')
    assert response.status_code == status.HTTP_400_BAD_REQUEST
```

### Everything else looks good

- `ban_user` service is correctly idempotent and uses `update_fields`
- `delete_tag` wrapping in `transaction.atomic()` with explicit `StoryTag` deletion ensures post_delete signals fire for story-count bookkeeping
- `UserBanResponseSerializer` exposes only the fields needed by the response
- Permission and 404 coverage in tests is thorough
- URL routing follows the existing `/moderation/` convention

KmalM (approved):
The one blocking item from the previous review is resolved — `test_missing_password_returns_400` is restored in `TestDeleteAccountView` (36/36 tests pass).

Everything else noted as good in the prior review still holds:
- `ban_user` is correctly idempotent with `update_fields=['is_active']`
- `delete_tag` wrapped in `transaction.atomic()` with explicit `StoryTag` deletion so `post_delete` signals fire for story-count bookkeeping
- `UserBanResponseSerializer` exposes only the necessary fields
- Permission/404 coverage is thorough on both endpoints
- URL routing follows the `/moderation/` convention

Approved.

PR #527, opened 2026-05-03
Feat: mobile timeline view

OsmanYusufTosun:
# Description
Addresses #190

Adds the mobile Timeline browsing experience so stories can be explored chronologically from the main app flow. This PR includes:

- Timeline as a main mobile pager route, bottom navigation item, and deep link target.
- A `TimelineScreen` with shared search/filter integration, refresh, loading, empty, error, and infinite-scroll states.
- Time-window selection for all stories, exact year, year range, and decade.
- Timeline cards that show historical timing, title, location, image previews, and open the story detail screen.
- Timeline service, repository, remote source, mapper, and domain entities for paginated timeline data.
- API support for `/stories/timeline/` with year bounds and map bounds, plus fallback fetching for query, location text, and proximity filters.
- `storyService.getTimeline()` so the timeline can be consumed through the existing story service layer.
- Focused tests for navigation, service delegation, remote source behavior, mappers, and timeline screen interactions.

Note: the request layer supports approximate periods, but this PR exposes the timeline picker UI for all/year/range/decade. A first-class approximate-period selector can be added as a follow-up if the full #190 selector scope is required before closing the issue.

# Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Documentation

# Acceptance Criteria / Checklist
- [x] Timeline screen is accessible from main mobile navigation
- [x] Story list is displayed chronologically from the timeline source
- [x] Tapping a timeline story opens story detail
- [x] Shared search and location filters are respected by the timeline screen
- [x] Loading, empty, error, refresh, and pagination states are handled
- [x] `storyService.getTimeline()` is added
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of the code
- [x] I have added/updated tests
- [x] Relevant unit tests pass locally with my changes

# Verification
- `npm --prefix mobile run typecheck`
- `npm --prefix mobile test -- --runInBand mobile/src/features/timeline/presentation/screens/__tests__/TimelineScreen.test.tsx mobile/src/features/timeline/data/mappers/__tests__/timelineMappers.test.ts mobile/src/features/timeline/data/sources/__tests__/timelineRemoteSource.test.ts mobile/src/app/navigation/__tests__/RootNavigator.test.tsx mobile/src/features/stories/application/services/__tests__/storyService.test.ts`

Test output includes a React `act(...)` warning from `MapCard` during the RootNavigator suite, but all 5 targeted suites passed: 37 tests total.

# Screenshots / Recordings
[...]

PR #538, opened 2026-05-06
feat: timeline view page with time-window filter

### LINKED ISSUE #183: [Feature][Frontend] Timeline view for exploring stories by time period with location scope
## Description

The platform needs a timeline-based feature for exploring stories by time period (requirements 1.3.4.1-1.3.4.4). This is the third major browsing interface alongside the map and feed. The timeline allows users to navigate history chronologically, optionally scoped to a specific location.

### What needs to be built

1. **Timeline page (`/timeline`):** A new route accessible from the main navigation bar. The page has two main sections: a time period selector at the top and a story list below.

2. **Time period selector:** An interactive control that lets users select:
   - A specific year (e.g., "1923")
   - A year range (e.g., "1914-1918")
   - A decade (e.g., "1920s")
   - An approximate period (e.g., "Early 1900s")

   This could be implemented as a visual timeline slider, a set of decade buttons, or a combination of input fields. The selector should be intuitive and visually represent the passage of time.

3. **Story list:** Below the selector, stories matching the selected time period are displayed in chronological order. Each story entry shows the same card format used in the feed (reuses `StoryCard` component). Clicking a story opens its detail page.

4. **Location scope (requirement 1.3.4.3):** The timeline can optionally be scoped to a specific location. This can be implemented as:
   - A location text input/filter above or beside the time selector
   - Integration with the existing location filter from the search/filter system
   - When a location is selected, only stories from that area appear on the timeline

5. **Integration with filters:** The timeline should respect and update the shared filter state (URL query params or context) so that navigating between timeline, feed, and map preserves filter context (requirement 1.3.5.5).

6. **Empty and loading states:** Use shared UI components from MVP for loading skeletons and empty states ("No stories found for this time period").

7. **API service:** Add `storyService.getTimeline({ year, yearRange, decade, location })` to the story service layer.

### Why this matters

The timeline feature is one of the three pillars of story discovery (map, feed, timeline) defined in the requirements (1.3.4). It enables a uniquely historical perspective — users can see how a neighborhood or city evolved over decades, which is the core mission of the platform.

## Acceptance Criteria

- [ ] Timeline page at route `/timeline`, accessible from main navigation
- [ ] Time period selector supporting: exact year, year range, decade, approximate period
- [ ] Stories displayed in chronological order using `StoryCard` component
- [ ] Clicking a story navigates to `/stories/:id`
- [ ] Location scope: optional location filter narrows stories to a specific area
- [ ] Timeline reflects active search/filter criteria from shared state
- [ ] Filter state preserved when navigating to a story and back
- [ ] Empty state when no stories match the selected period
- [ ] Loading state while fetching stories
- [ ] Pagination or infinite scroll for large result sets
- [ ] API service: `storyService.getTimeline()`
- [ ] Navigation bar updated to include Timeline link
- [ ] Unit tests: time selector interactions, story list rendering, location scope, filter integration, empty/loading states

omarreis01:
# Description
Fixes #183

Adds a public Timeline page that lists all stories chronologically (oldest first) using the existing `GET /stories/timeline/` backend endpoint. Mirrors the mobile UX (PR #527): a "Choose a time window" filter card with **All / Year / Range / Decade** modes, a period label + story count, and a vertical timeline with year-bullet markers and story cards (image variant when `photo_url` is present, plain card otherwise). Each entry links to the corresponding story detail page.

The service layer also accepts bounding-box params via the URL query string, so issue #488 ("View Timeline" map-pin action) can reuse this page directly without modification — it will pass `lat_min/lat_max/lng_min/lng_max` derived from the pin's coordinates.

**Highlights**
- New service `timelineService.js` wrapping the timeline endpoint (camelCase → snake_case, omits undefined params, bbox passthrough).
- New page `TimelinePage.jsx` with filter modes, "Load more" pagination, and empty/loading/error states.
- New `Timeline/TimelineView.jsx` and `Timeline/TimelineEntry.jsx` components — vertical line, year bullet, EDTF/decade chips, "Mapped at lat, lng" hint, clickable card.
- Route `/timeline` registered in `App.jsx`; "Timeline" nav link added to `AppLayout` between Map and Feed (matches mobile nav order).
- Stale-response guard via generation counter so rapid filter changes don't render outdated data.
- Range mode validates `from <= to` inline and skips the fetch until the range is valid.

# Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Documentation

# Screenshots / Recordings

# Estimated Review Time
- [ ] < 5 min
- [x] 5-15 min
- [ ] > 15 min

### DIFF
--- a/frontend/src/App.jsx
+++ b/frontend/src/App.jsx
@@ -14,6 +14,7 @@ import ProfileCompletionPage from "@/pages/ProfileCompletionPage";
 import NotFoundPage from "@/pages/NotFoundPage";
 import FeedPage from "@/pages/FeedPage";
 import MapPage from "@/pages/MapPage";
+import TimelinePage from "@/pages/TimelinePage";
 import ProfilePage from "@/pages/ProfilePage";
 import SubmitStoryPage from "@/pages/SubmitStoryPage";
 import StoryDetailPage from "@/pages/StoryDetailPage";
@@ -28,6 +29,7 @@ function App() {
             <Route element={<AppLayout />}>
               <Route path="/" element={<FeedPage />} />
               <Route path="/map" element={<MapPage />} />
+              <Route path="/timeline" element={<TimelinePage />} />
               <Route path="/login" element={<LoginPage />} />
               <Route path="/register" element={<RegisterPage />} />
               <Route path="/forgot-password" element={<ForgotPasswordPage />} />
--- a/frontend/src/components/AppLayout/AppLayout.jsx
+++ b/frontend/src/components/AppLayout/AppLayout.jsx
@@ -15,6 +15,7 @@ import { useAuth } from "@/hooks/useAuth";
 
 const publicLinks = [
   { to: "/map", label: "Map", preserveSearch: true },
+  { to: "/timeline", label: "Timeline", preserveSearch: true },
   { to: "/", label: "Feed", preserveSearch: true },
   { to: "/submit-story", label: "Submit Story", icon: Plus },
 ];
--- a/frontend/src/components/Timeline/TimelineEntry.jsx
+++ b/frontend/src/components/Timeline/TimelineEntry.jsx
@@ -1,0 +1,106 @@
+import { Link } from "react-router-dom";
+import { MapPin } from "lucide-react";
+
+import { cn } from "@/lib/utils";
+import { formatTimePeriod } from "@/components/StoryCard/storyCardUtils";
+
+/**
+ * Best-effort label for the bullet on the timeline. Prefers the formatted
+ * time period (year, "1870s", c. 1875, etc.); falls back to year_start
+ * or the leading 4 digits of date_value when present.
+ */
+function bulletLabel(story) {
+  const formatted = formatTimePeriod(story);
+  if (formatted) return formatted;
+  if (story.year != null) return String(story.year);
+  if (story.year_start != null) return String(story.year_start);
+  if (typeof story.date_value === "string") {
+    const match = story.date_value.match(/-?\d{4}/);
+    if (match) return match[0];
+  }
+  return "";
+}
+
+function decadeChip(story) {
+  const yearForDecade = story.year ?? story.year_start;
+  if (yearForDecade == null) return null;
+  return `${Math.floor(yearForDecade / 10) * 10}s`;
+}
+
+function TimelineEntry({ story }) {
+  const label = bulletLabel(story);
+  const decade = story.time_type === "decade" ? null : decadeChip(story);
+  const hasCoords =
+    story.location_lat != null && story.location_lng != null;
+  const lat = hasCoords ? Number(story.location_lat).toFixed(4) : null;
+  const lng = hasCoords ? Number(story.location_lng).toFixed(4) : null;
+
+  return (
+    <div className="relative flex gap-4 pl-2">
+      {/* Bullet */}
+      <div className="relative flex flex-col items-center" aria-hidden="false">
+        <div
+          className={cn(
+            "z-10 flex min-w-[3.5rem] items-center justify-center rounded-full border bg-background px-2 py-1",
+            "text-xs font-medium text-foreground shadow-sm",
+          )}
+        >
+          {label}
+        </div>
+      </div>
+
+      {/* Card */}
+      <Link
+        to={`/stories/${story.id}`}
+        className={cn(
+          "group flex-1 overflow-hidden rounded-xl border bg-card text-card-foreground shadow-sm",
+          "transition-shadow hover:shadow-md focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-ring",
+        )}
+        aria-label={`Read story: ${story.title}`}
+      >
+        {story.photo_url && (
+          <img
+            src={story.photo_url}
+            alt={story.title}
+            className="h-40 w-full object-cover"
+            loading="lazy"
+          />
+        )}
+        <div className="space-y-2 p-4">
+          <h3 className="text-base font-semibold leading-snug text-foreground line-clamp-2">
+            {story.title}
+          </h3>
+
+          <div className="flex flex-wrap gap-1.5">
+            {label && (
+              <span className="inline-flex items-center rounded-full bg-muted px-2 py-0.5 text-xs text-muted-foreground">
+                {label}
+              </span>
+            )}
+            {decade && decade !== label && (
+              <span className="inline-flex items-center rounded-full bg-muted px-2 py-0.5 text-xs text-muted-foreground">
+                {decade}
+              </span>
+            )}
+            {story.temporal_coverage && (
+              <span className="inline-flex items-center rounded-full bg-muted px-2 py-0.5 text-xs text-muted-foreground">
+                {story.temporal_coverage}
+              </span>
+            )}
+          </div>
+
+          {hasCoords && (
+            <div className="flex items-center gap-1.5 text-xs text-muted-foreground">
+              <MapPin className="h-3.5 w-3.5 shrink-0" aria-hidden="true" />
+              <span>
+                Mapped at {lat}, {lng}
+              </span>
+            </div>
+          )}
+        </div>
+      </Link>
+    </div>
+  );
+}
+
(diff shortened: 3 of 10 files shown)


sezindogan:
## Overview                                                                                                                                             
                                                                                                                                                       
  Adds a public /timeline route to the frontend: a time-window filter card (All / Year / Range / Decade) + vertical timeline with year-bullet markers and story cards. The service layer supports bbox passthrough for the future map-pin integration (#488). The PR claims to mirror the mobile UX (PR #527).                                                                                                                                               
                            
##  Strengths
  - Generation counter stale-response guard is well-implemented and matches the pattern used on mobile.                                                
  - deriveYearFilter cleanly separates the filter logic from the render tree.                                                                          
  - timelineService.js KEY_MAP is clean — camelCase→snake_case in one place, easy to maintain.                                                         
  - Test coverage is solid: service params, page behavior, component rendering, load-more flow, bbox passthrough.                                      
                                                                                                                                                       
 ## Issues                                                                                                                                               
                                                                                                                                                       
 ### Behavioral divergences from mobile (important)                                                                                                       
                            
  1. Partial year inputs fire requests. Mobile's TimelinePeriodSelector validates with /^\d{4}$/ — it only requests when the user has typed a complete 4-digit year, and clamps input to 4 digits. Frontend's parseIntOrUndef accepts any integer, so entering "1875" fires four consecutive requests (year=1, year=18, year=187, year=1875). A debounce or the same 4-digit guard would fix this.                                                         
  2. Range mode allows one-sided ranges. Mobile requires both rangeFrom and rangeTo to be complete 4-digit years before fetching (shows "Enter start and end years" otherwise). Frontend's deriveYearFilter at line 460–463 falls through to set yearFrom or yearTo alone when only one side is filled — this fires a fetch with a half-baked range.
  3. Coordinates shown instead of location name. TimelineEntry shows Mapped at 41.0258, 28.9744 (raw lat/lng, 4 dp). Mobile's TimelineCard shows story.locationName — a human-readable place name. If the API response includes a location_name field, it should be preferred; raw coordinates are not user-friendly.
                                                                                                                                                       
 ### Accessibility                                                                                                                                        
   
  4. aria-hidden="false" is redundant (TimelineEntry.jsx:79). false is the default; remove it.                                                         
  5. Semantically incorrect ARIA role for the mode selector. The four mode pills are mutually exclusive (only one can be active), which maps to a radiogroup/radio pattern. The current role="group" + aria-pressed treats them as independent toggles. Mobile correctly uses accessibilityRole="radiogroup" and accessibilityRole="radio". Fix: <div role="radiogroup" aria-label="Time window mode" ...>  <button role="radio" aria-checked={mode === m} ...>                                                                                                
   
 ### Minor                                                                                                                                                
                            
  6. Dead code in timelineService.js. hasImage is in KEY_MAP and the function signature but is never passed from TimelinePage. Either wire it up or remove it.                                                                                                                                                             

##  Missing Features — Mobile Parity Gaps                                                                                                                
                                       
  The following are present in mobile's TimelineScreen but absent from this PR. They should be addressed before merge (or explicitly deferred with a follow-up issue).                                                                                                                                    
   
####  Search bar (title/text search)                                                                                                                       
                            
  Mobile renders <StorySearchControls helperText="Search by title or place." /> above the period selector. The frontend has an equivalent SearchFilter / SearchBar component used on the Feed page and Map pages but it not implemented it timeline. The timeline page has no text search at all. 
                                                                                                                                                       
#### Location filter (Nominatim geocoding)                                                                                                                
   
  Mobile geocodes the typed place name via Nominatim → resolves to a LocationBounds bbox → passes lat_min/lat_max/lng_min/lng_max to the timeline endpoint. The timelineService.js already has bbox passthrough wired up (the lat_min/lat_max/lng_min/lng_max params exist), but there is no UI to populate them from user input. 

  Note: A separate PR #531 adding Nominatim geocoding to the frontend FilterPanel is pending. Once that lands, this page should plug into it — the bbox output from that geocoder maps directly to the existing passthrough params in timelineService.js. This is a hard blocker for location filtering parity; the timeline page should not be considered complete without it.                                                                              
                            
####  Tag filter

  Mobile's filter panel includes tag multi-select (with live backend search). The frontend already has TagFilterInput.jsx and the tag search service   
  used on the Feed page — these should be wired into the timeline page's filter state, and the getTimeline service call needs a tags param added.
                                                                                                                                                       
 #### Active filter chips       

  Mobile displays active filters as dismissible chips below the search bar (location, time range, tags, etc.). The frontend has ActiveFilters.jsx for  this, used on the Feed page. The timeline page shows no visual indication of which filters are active beyond the period label.

### Summary
Essentially, the search and filter functionalities of the map and feed views should be implemented in the timeline view as well.

omarreis01:
Thanks for the thorough review @sezindogan! Pushed `6ed8094` addressing all 6 items in the **Issues** section.

### Fixed

1. **Partial year inputs no longer fire requests.** Added a `/^\d{4}$/` guard in `parseFullYear` — typing `1875` now fires a single request once the 4th digit is entered, instead of firing for `1`, `18`, `187`, `1875`. Same guard applies to year and decade modes.
2. **Range mode requires both endpoints.** `deriveYearFilter` no longer falls through with only one side filled — it returns `skip: true` and the status row shows *"Enter start and end years"* until both inputs are complete 4-digit years (and `from <= to`).
3. **Location name preferred over raw coordinates.** `TimelineEntry` now uses `story.location_name` when present and only falls back to `Mapped at lat, lng` if no name is provided. *(Note: the current `StoryTimelineSerializer` at `backend/apps/stories/serializers.py:481` doesn't include `location_name` yet — flagging for awareness; the frontend will pick it up automatically once it's added on the backend side.)*
4. **Removed redundant `aria-hidden="false"`** from the bullet wrapper in `TimelineEntry.jsx`.
5. **Mode selector now uses the correct ARIA pattern.** Switched from `role="group"` + `aria-pressed` to `role="radiogroup"` + `role="radio"` + `aria-checked`, matching the mobile `accessibilityRole="radiogroup"` semantics.
6. **Dropped dead `hasImage` code** from `timelineService.js` (key map, function signature, and corresponding test assertion).

### Tests
Updated existing tests to match the new ARIA roles, added three new tests covering the guards (partial year doesn't fetch, one-sided range doesn't fetch, full range does fetch), and added a `TimelineEntry` test for the `location_name` preference. All 28 timeline tests pass; full suite + lint + build are clean.

### Mobile parity gaps (search / location / tags / active chips)
Agreed these belong on the timeline page. Since PR #531 will land the Nominatim geocoding integration that this page should plug into, I'd like to handle the search/location/tag/active-chip integration in a follow-up PR once #531 merges (so the location filter can use the same geocoder rather than duplicating it). Happy to open a tracking issue if you'd prefer.

sezindogan (approved):
Items 1–6: all confirmed  
                                                                                                                                                       
  1. parseFullYear with /^\d{4}$/ guard is in place; deriveYearFilter propagates skip: true for partial input, and the useEffect bails on filter.skip. 
  Year and decade modes both use it.                                                                                                                   
  2. Range mode returns skip: true with "Enter start and end years" when either side is missing or partial. The edge case where both are empty         
  correctly falls through to "All time periods" without setting skip.                                                                                  
  3. location_name preference is correctly implemented — falls back to Mapped at lat, lng only when location_name is absent or whitespace-only.
  4. aria-hidden="false" removed from the bullet wrapper. ✓                                                                                            
  5. role="radiogroup" + role="radio" + aria-checked in place. ✓                                                                                       
  6. hasImage dropped from KEY_MAP, function signature, and args object in the service. ✓                                                              
                                                                                                                                                       
  One small note on the ARIA radiogroup (non-blocking): The WAI-ARIA authoring spec for radiogroup expects a roving tabindex — only the checked radio in the tab order, arrow keys to move between options. Right now each pill is independently tab-focusable, which deviates from the pattern.           
  Functionally fine for mouse/touch users, but worth a quick follow-up for full keyboard accessibility.                                                
                            
  Tests: The three new guard tests and the location_name preference test all look well-scoped and cover the real behavior.                             
   
  On the follow-up: please do open a tracking issue for the search / location / tag / active-chip integration so it doesn't get lost.

PR #540, opened 2026-05-06
feat: proxy Nominatim geocoding through Django 

### LINKED ISSUE #534: [Feature][Backend] Proxy Nominatim geocoding requests through Django
### Description

                                                                                                                                                     
  Currently, both the frontend and mobile clients call nominatim.openstreetmap.org directly from the browser/app for location autocomplete. This violates two requirements of Nominatim's usage policy:                                                                                                            
                                                                                                                                                       
  1. Identification (User-Agent)                                      
  Nominatim requires every client to identify itself with a meaningful User-Agent header. Both clients attempt to set this:
  - Frontend: LocalHistoryStoryMap/1.0                                                                                                                 
  - Mobile: local-history-story-map-mobile/0.1.0                                                                                                       
                                                                                                                                                       
  However, User-Agent is a forbidden header name — both browsers and React Native's fetch silently drop it. Nominatim never sees it. Unidentified requests are grounds for blocking under their policy.                                                                                                
   
  2. Rate limit                                                                                                                                        
  Nominatim caps usage at 1 req/s per IP. With a 250ms debounce on the frontend and multiple users sharing the same egress IP once deployed, this limit can be exceeded. Hitting it returns a 429, which currently surfaces as the suggestions dropdown silently disappearing with no feedback to the user. 
   
  Best solution: proxy through Django                                                                                                                  
                                                                      
  A Django proxy endpoint can set a real User-Agent via Python's requests library, enforce the rate limit server-side, and cache responses so repeated queries never reach Nominatim at all. This fixes both issues for both clients in one place.
                                                                                                                                                       
  Proposed endpoints:                                                 
  - GET /api/geocode/?q=Istanbul — resolves a place name to a bounding box; uses Nominatim /search                                                     
  - GET /api/geocode/suggestions/?q=Istanbul — returns up to 5 autocomplete suggestions with bbox; used by the filter panel and story submission map
  - GET /api/geocode/reverse/?lat=41.0&lng=28.9 — resolves a dropped pin to a place name; uses Nominatim /reverse, useful for story submission map     
                                                               
                                                                                     
  Response shape should match what the frontend's geocodingService.js and mobile's geocodingService.ts currently expect so the client-side change is a one-line URL swap in each.                                                                                                                           

### Acceptance Criteria

  - [ ] GET /api/geocode/ returns { lat_min, lat_max, lng_min, lng_max } or null                                                                           
  - [ ] GET /api/geocode/suggestions/ returns an array of { id, title, subtitle, bbox }
  - [ ] GET /api/geocode/reverse/ returns a place name string or null                                                                                      
  - [ ] All three endpoints proxy to Nominatim with a valid User-Agent identifying the app
  - [ ] Outgoing requests to Nominatim are rate-limited to ≤ 1 req/s server-side                                                                           
  - [ ] Responses are cached (Django cache framework, suggested TTL: 24h)                                                                                  
  - [ ] Frontend geocodingService.js switches NOMINATIM_BASE to the new backend endpoint after this issue is implemented.                                                                  
  - [ ] Mobile geocodingService.ts switches NOMINATIM_SEARCH_URL to the new backend endpoint after this issue is implemented.                                                               


### Meta (optional)

  Effort: M                                                           
  Priority: P2
  Deadline: ASAP

KmalM:
# 📝 Description
Fixes #534

Proxies all Nominatim geocoding requests through Django instead of hitting `nominatim.openstreetmap.org` directly from the browser/app. This fixes two Nominatim usage-policy violations:

1. **User-Agent** — browsers/React Native silently drop `User-Agent` headers; Python's `requests` sends it correctly.
2. **Rate limit** — a server-side threading lock enforces ≤ 1 req/s; 24 h response caching means most repeated queries never reach Nominatim at all.

**New endpoints:**

| Endpoint | Returns |
|---|---|
| `GET /api/geocode/?q=Istanbul` | `{ lat_min, lat_max, lng_min, lng_max }` or `null` |
| `GET /api/geocode/suggestions/?q=Ist` | `[{ id, title, subtitle, bbox }]` (empty if query < 3 chars) |
| `GET /api/geocode/reverse/?lat=41.0&lng=28.9` | `{ place_name: "..." }` or `{ place_name: null }` |

**Implementation notes:**
- `apps/geocoding` is a pure-proxy app — no Django models.
- Query strings are lowercased before being used as cache keys so `Istanbul` and `ISTANBUL` share the same cache entry.
- Coordinates are validated (numeric, within range) before forwarding to Nominatim; invalid inputs return `null` without an upstream call.
- `NominatimUnavailable` (HTTP 503) is raised as a DRF exception when Nominatim is unreachable or returns a 5xx, so errors propagate cleanly through the DRF exception handler.
- `GeocodeView` returns `JsonResponse(result, safe=False)` rather than `Response(result)` because DRF's `JSONRenderer.render(None)` returns `b''` (empty body), not `b'null'` — verified empirically.
- The module-level rate-limiter works within a single process; a note in the code flags the multi-worker gunicorn limitation.

# 📊 Type of Change
- [x] New feature

- [x] `GET /api/geocode/` returns `{ lat_min, lat_max, lng_min, lng_max }` or null
- [x] `GET /api/geocode/suggestions/` returns an array of `{ id, title, subtitle, bbox }`
- [x] `GET /api/geocode/reverse/` returns a place name string or null
- [x] All three endpoints proxy to Nominatim with a valid User-Agent identifying the app
- [x] Outgoing requests to Nominatim are rate-limited to ≤ 1 req/s server-side
- [x] Responses are cached (Django cache framework, 24h TTL)
- [x] New and existing unit tests pass locally with my changes (1558 passed)

# ⏱️ Estimated Review Time
- [ ] < 5 min
- [x] 5-15 min
- [ ] > 15 min

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -33,3 +33,8 @@ DEFAULT_FROM_EMAIL=Local History Story Map <noreply@yourdomain.com>
 
 # Base URL of the frontend — used to build password reset links in emails.
 FRONTEND_URL=http://localhost:3000
+
+# ── Nominatim geocoding proxy ─────────────────────────────────────────────────
+# Nominatim policy requires a real contact address in the User-Agent header.
+# Set this to the project maintainer's email for staging/production deployments.
+NOMINATIM_CONTACT_EMAIL=admin@example.com
--- a/backend/apps/geocoding/apps.py
+++ b/backend/apps/geocoding/apps.py
@@ -1,0 +1,6 @@
+from django.apps import AppConfig
+
+
+class GeocodingConfig(AppConfig):
+    default_auto_field = 'django.db.models.BigAutoField'
+    name = 'apps.geocoding'
--- a/backend/apps/geocoding/services.py
+++ b/backend/apps/geocoding/services.py
@@ -1,0 +1,131 @@
+import hashlib
+import threading
+import time
+
+import requests
+from django.core.cache import cache
+
+_NOMINATIM_BASE = 'https://nominatim.openstreetmap.org'
+_USER_AGENT = 'LocalHistoryStoryMap/1.0 (Django proxy; contact: admin@example.com)'
+_CACHE_TTL = 86400  # 24 hours
+_SENTINEL = object()  # distinguishes "cached as None/[]" from "cache miss"
+
+_rate_lock = threading.Lock()
+_last_call_time = 0.0
+
+
+def _nominatim_get(path, params):
+    """Call Nominatim, honouring the 1 req/s server-side rate limit."""
+    global _last_call_time
+    with _rate_lock:
+        now = time.monotonic()
+        gap = now - _last_call_time
+        if gap < 1.0:
+            time.sleep(1.0 - gap)
+        _last_call_time = time.monotonic()
+
+    response = requests.get(
+        f'{_NOMINATIM_BASE}{path}',
+        params=params,
+        headers={
+            'User-Agent': _USER_AGENT,
+            'Accept': 'application/json',
+            'Accept-Language': 'en',
+        },
+        timeout=10,
+    )
+    response.raise_for_status()
+    return response.json()
+
+
+def _cache_key(endpoint, raw_params):
+    fingerprint = hashlib.md5(f'{endpoint}:{sorted(raw_params.items())}'.encode()).hexdigest()
+    return f'geocode:{endpoint}:{fingerprint}'
+
+
+def _parse_bbox(boundingbox):
+    """Convert Nominatim boundingbox list to dict, ensuring min < max."""
+    if not isinstance(boundingbox, (list, tuple)) or len(boundingbox) < 4:
+        return None
+    try:
+        values = [float(v) for v in boundingbox[:4]]
+    except (ValueError, TypeError):
+        return None
+    lat_min, lat_max = min(values[0], values[1]), max(values[0], values[1])
+    lng_min, lng_max = min(values[2], values[3]), max(values[2], values[3])
+    return {'lat_min': lat_min, 'lat_max': lat_max, 'lng_min': lng_min, 'lng_max': lng_max}
+
+
+def geocode_search(q):
+    """Return { lat_min, lat_max, lng_min, lng_max } for q, or None."""
+    q = q.strip()
+    if not q:
+        return None
+
+    params = {'q': q, 'format': 'jsonv2', 'limit': '1', 'addressdetails': '1'}
+    key = _cache_key('search', params)
+    cached = cache.get(key, _SENTINEL)
+    if cached is not _SENTINEL:
+        return cached
+
+    results = _nominatim_get('/search', params)
+    if not isinstance(results, list) or not results:
+        result = None
+    else:
+        result = _parse_bbox(results[0].get('boundingbox'))
+
+    cache.set(key, result, _CACHE_TTL)
+    return result
+
+
+def geocode_suggestions(q):
+    """Return up to 5 { id, title, subtitle, bbox } suggestions for q."""
+    q = q.strip()
+    if len(q) < 3:
+        return []
+
+    params = {'q': q, 'format': 'jsonv2', 'limit': '5', 'addressdetails': '1'}
+    key = _cache_key('suggestions', params)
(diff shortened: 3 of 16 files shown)


AysuKeskin (changes requested):
Overall, the implementation looks solid. The sentinel cache pattern, bbox normalization, coordinate clamping, and the comment about rate-lock limitations in multi-worker deployments are all good choices.

### Issues

**1. Missing `test_services.py`**

The public service functions `geocode_search`, `geocode_suggestions`, and `geocode_reverse` are currently only covered through view-level tests. Since the core logic lives in `services.py`, we should add isolated service tests as well.



**2. Placeholder contact email in `_USER_AGENT`**

`admin@example.com` should not be used as the Nominatim contact email:

```python
_USER_AGENT = 'LocalHistoryStoryMap/1.0 (Django proxy; contact: admin@example.com)'
````

Nominatim requires a real, reachable contact address. This should either be replaced with a real project email or made configurable through settings, e.g. `settings.NOMINATIM_CONTACT_EMAIL`.

**3. `JsonResponse` usage is inconsistent with the rest of the DRF views**

`GeocodeView` uses `JsonResponse`, while the other views use DRF `Response`. It would be cleaner to keep this consistent and use:

```python
return Response(result)
```

DRF should serialize `None` as JSON `null` with the default JSON renderer. If `JsonResponse` is still needed for a specific reason, that behavior should be documented clearly.

**4. URL prefix convention**

The new endpoint is mounted under `api/geocode/`, while the existing project routes use prefixes like `auth/`, `users/`, `stories/`, and `tags/` without `api/`.

For consistency, this should probably be changed to `geocode/`, unless the whole project is intentionally moving toward an `api/` prefix.

**5. Unused `addressdetails=1` in `geocode_search`**

`geocode_search` fetches `addressdetails`, but only returns the bounding box. This adds unnecessary payload size:

```python
params = {'q': q, 'format': 'jsonv2', 'limit': '1', 'addressdetails': '1'}
```

I’d remove `addressdetails` from `geocode_search` and keep it only where address parts are actually used, such as suggestions.

### Summary

The core logic is well-written, especially `_nominatim_get`, `_parse_bbox`, caching, and coordinate validation. Before merging, I think the main blockers are adding isolated service tests and replacing the placeholder Nominatim contact email. The URL prefix inconsistency and unused `addressdetails` parameter are also worth fixing in this PR.

KmalM:
Items 1, 2, 4, and 5 are addressed in the latest commits. Item 3 (JsonResponse) is intentional — here's why.

**Item 1 (service tests)** — Added `test_services.py` with 31 isolated tests covering `_parse_bbox`, the cache sentinel pattern, coordinate normalisation for cache keys, and all three public service functions. 56 total tests pass.

**Item 2 (contact email)** — `_USER_AGENT` now reads from `settings.NOMINATIM_CONTACT_EMAIL`, configured via env var. Added to `base.py` with `env('NOMINATIM_CONTACT_EMAIL', default='admin@example.com')` and documented in `.env.example`.

**Item 3 (`JsonResponse` vs `Response`)** — Keeping `JsonResponse`. DRF's `JSONRenderer.render(None)` returns `b''` (empty bytes), **not** `b'null'`. A client calling `JSON.parse('')` or `.json()` raises a parse error. `JsonResponse(None, safe=False)` is the correct way to emit `null` here. This was verified empirically and is documented in the PR body — the inconsistency is real but the alternative is broken behaviour.

**Item 4 (URL prefix)** — Changed `api/geocode/` → `geocode/` in `config/urls.py` to match every other app. Test URL constants updated accordingly.

**Item 5 (`addressdetails` in `geocode_search`)** — Removed. Only `boundingbox` is used from the search result; `addressdetails` was unnecessary payload.

AysuKeskin (approved):
Thanks for addressing the review comments.

The latest changes resolve the main blockers I mentioned earlier: service-level tests were added, the Nominatim contact email is now configurable, the URL prefix is consistent with the rest of the project, and the unnecessary `addressdetails` parameter was removed from `geocode_search`.

Overall, the implementation looks good to me. The caching, rate limiting, coordinate validation, and error handling are well covered. Approving this PR.

PR #542, opened 2026-05-06
feat: Add a public endpoint for listing a user's published stories

### LINKED ISSUE #541: [Feature][Backend] Add paginated published-stories endpoint for public user profiles
### Description

The current public profile endpoint, `GET /users/<user_id>/`, returns profile information and `published_story_count`, but it does not return the user’s story list.

Instead of embedding all stories directly into the profile response, the backend should provide a separate **paginated endpoint** for listing a user’s published stories. This is the cleaner approach for the current architecture because returning all stories inside the profile response would make the payload unnecessarily large and would not scale well. A separate endpoint is also more consistent with the existing design, where similar list-style data such as bookmarks is already handled through dedicated endpoints.

The proposed endpoint is:

`GET /users/<user_id>/stories/`

This endpoint should return only the user’s **published** stories. Draft and removed stories must not be exposed through the public profile story list. If the target user is inactive or banned, the endpoint should return `404 Not Found`.

The response should be paginated and can reuse the existing `StoryFeedSerializer` so the returned story cards stay consistent with the rest of the application. If the request is authenticated, the queryset should also support the existing `user_has_liked` and `user_has_saved` annotations so the frontend gets the same interaction-aware response shape it already expects in story feed-style endpoints.

### Acceptance Criteria

* [ ] A new endpoint `GET /users/<user_id>/stories/` is added
* [ ] The endpoint returns only published stories
* [ ] Draft and removed stories are excluded
* [ ] Inactive or banned users return `404 Not Found`
* [ ] The response is paginated
* [ ] The endpoint reuses `StoryFeedSerializer` or an equivalent feed-compatible response shape
* [ ] If the requester is authenticated, `user_has_liked` and `user_has_saved` annotations are included correctly
* [ ] Stories are ordered by `-submitted_at`
* [ ] Tests are added or updated for the endpoint behavior
* [ ] New and existing relevant tests pass

### Meta (optional)

Estimated Duration: 2–3 hours

AysuKeskin:
# 📝 Description

Fixes #541

This PR adds a new public and paginated endpoint for listing the published stories of a specific user:

`GET /users/<user_id>/stories/`

The purpose of this PR is to separate public profile information from the user's story list. Instead of embedding all stories directly inside the public profile response, this endpoint provides a dedicated and paginated way to fetch a user's published stories. This keeps the profile response lightweight and makes the story list easier to load, paginate, and extend on the frontend.

This endpoint is intended to be used on public profile pages where visitors need to see the stories shared by a user. It only returns stories that are publicly visible, so draft and removed stories are excluded from the response. If the requested user does not exist or the account is inactive/banned, the endpoint returns `404`.

Implemented changes:
- Added `UserStoriesView` to handle public profile story listing.
- Added the `get_user_published_stories(user_id)` service function to keep the queryset logic outside the view.
- Registered the new `/users/<user_id>/stories/` route in `profile_urls.py`.
- Reused the existing `StoryFeedSerializer` so the response format stays consistent with the story feed.
- Reused `StoryPagination` so the endpoint returns paginated results.
- Filtered the queryset to return only published stories.
- Excluded draft and removed stories from the public response.
- Added `404` behavior for nonexistent users and inactive/banned users.
- Allowed unauthenticated users to access the endpoint.
- Added authenticated interaction annotations so `user_has_liked` and `user_has_saved` are returned correctly when the requester is logged in.
- Ordered stories by `submitted_at` in descending order, so the most recently submitted stories appear first.
- Added service and view tests for filtering, pagination, unauthenticated access, interaction flags, empty results, ordering, and `404` cases.

# 📊 Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Documentation

- [x] Style guidelines followed
- [x] Code commented where necessary
- [x] Self-review performed
- [x] Tests added/updated
- [x] Tests pass locally

# ⏱️ Estimated Review Time

- [ ] < 5 min
- [x] 5–15 min
- [ ] > 15 min

### DIFF
--- a/backend/apps/users/profile_urls.py
+++ b/backend/apps/users/profile_urls.py
@@ -8,6 +8,7 @@ from apps.users.views import (
     ProfilePhotoView,
     UserBookmarksView,
     UserPublicProfileView,
+    UserStoriesView,
 )
 
 app_name = 'users_profile'
@@ -23,4 +24,5 @@ urlpatterns = [
     path('<int:user_id>/followers/', FollowerListView.as_view(), name='user-followers'),
     path('<int:user_id>/following/', FollowingListView.as_view(), name='user-following'),
     path('<int:user_id>/bookmarks/', UserBookmarksView.as_view(), name='user-bookmarks'),
+    path('<int:user_id>/stories/', UserStoriesView.as_view(), name='user-stories'),
 ]
--- a/backend/apps/users/services.py
+++ b/backend/apps/users/services.py
@@ -434,6 +434,29 @@ def get_user_bookmarks(user_id: int, requesting_user):
     )
 
 
+def get_user_published_stories(user_id: int):
+    """
+    Return a queryset of published stories for the given user's public profile,
+    ordered most-recently-submitted first.
+
+    Raises Http404 if the user does not exist or is inactive (banned).
+    """
+    from apps.stories.models import Story  # local import to avoid circular imports
+
+    try:
+        target = User.objects.get(pk=user_id, is_active=True)
+    except User.DoesNotExist:
+        raise Http404
+
+    return (
+        Story.objects
+        .filter(user=target, status=Story.STATUS_PUBLISHED)
+        .select_related('user', 'user__profile')
+        .prefetch_related('tags')
+        .order_by('-submitted_at')
+    )
+
+
 def request_password_reset(email: str) -> None:
     """
     Generates a password reset token and emails the reset link to the given address.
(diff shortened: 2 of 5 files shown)


KmalM (approved):
16/16 new tests pass locally. Implementation is clean — service/view split is correct, / are in place,  is applied only for authenticated requests (matching the  pattern), and 404 handling for inactive/missing users is correct. No blocking issues.

KmalM (approved):
16/16 new tests pass locally. Implementation is clean — service/view split is correct, `select_related`/`prefetch_related` are in place, `annotate_user_interactions` is applied only for authenticated requests (matching the `UserBookmarksView` pattern), and 404 handling for inactive/missing users is correct. No blocking issues.

PR #546, opened 2026-05-06
Fix/ expose comment ownership flag for frontend

### LINKED ISSUE #544: [Bug][Backend] Profile anonymization prevents comment deletion
### Description

A serialization issue prevents profile anonymization from working end-to-end. When a comment is anonymized or the author has a private profile, CommentResponseSerializer hides the author_username field by returning null.

Because the frontend identifies ownership by checking user.username === comment.author_username, it fails to recognize the current user's own anonymous comments. As a result, the delete button is never rendered, making it impossible to delete the comment while the profile is anonymized. If the user reverts their profile back to public, the button reappears and the comment can be deleted once again.

### How to Replicate

- Log into the application and navigate to a story.

- Post a comment on the story.

- Toggle your profile settings to Anonymous (or set it to private).

- Return to the story and view your comment.

- Observe: The delete button/option for your comment has disappeared.

- Toggle your profile settings back to Public.

- Return to the story.

- Observe: The delete button is visible again, and the comment can be deleted successfully.

### Current Behaviour

- When a comment is anonymized, the API payload returns "author_username": null.

- The frontend evaluation (user.username === comment.author_username) resolves to false.

- The frontend hides the delete action UI elements for the user's own anonymous comments.

### Expected Behaviour

- The frontend should be able to reliably identify ownership of a comment regardless of its anonymization or privacy state.

- The delete button should remain available to the author when viewing their own anonymous comments.                                                                                                                                    
                  
 ### Fix
 Add an is_own boolean field to CommentResponseSerializer using SerializerMethodField that returns True when request.user == comment.author. Pass context={'request': request} from the view.                                                                        
                 
                                                                              

### Acceptance Criteria

- [ ] GET /stories/<id>/comments/ includes is_own: true for comments authored by the authenticated requester, regardless of anonymization state.

- [ ] GET /stories/<id>/comments/ includes is_own: false for comments authored by other users.

- [ ] Existing serializer tests continue to pass.

- [ ] New unit tests are added covering both cases (anonymous and public ownership evaluation).

### Meta (optional)

_No response_

AysuKeskin:
# 📝 Description

Fixes #544

This PR fixes the problem where the frontend could not reliably determine whether a comment belongs to the currently authenticated user when the comment author's username is hidden.

Previously, comment responses only included `author_username`. However, `author_username` can be `null` when:
- the comment is anonymized,
- the author has a private username,
- or the author information should not be publicly displayed.

Because of this, the frontend could not safely decide whether to show owner-only actions such as edit/delete controls for a comment.

To solve this, this PR adds a new boolean field to comment responses:

`is_own_comment`

This field is independent from username visibility. It returns `true` only when the requesting authenticated user is the author of the comment, and `false` otherwise.

Implemented changes:
- Added `is_own_comment` to `CommentResponseSerializer`.
- Updated the comment list endpoint to pass `request` into serializer context.
- Updated the comment creation response to also include the correct `is_own_comment` value.
- Kept existing privacy behavior unchanged:
  - `author_username` is still hidden when the comment is anonymized or the author's username is private.
  - `is_own_comment` still works correctly even when `author_username` is `null`.
- Added serializer and view tests for authenticated users, unauthenticated users, other users, missing request context, comment creation, comment listing, and the private username regression case.

# 📊 Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactoring
- [ ] Documentation

- [x] Style guidelines followed
- [x] Code commented where necessary
- [x] Self-review performed
- [x] Tests added/updated
- [x] Tests pass locally

# ⏱️ Estimated Review Time

- [x] < 5 min
- [ ] 5–15 min
- [ ] > 15 min

### DIFF
--- a/backend/apps/interactions/serializers.py
+++ b/backend/apps/interactions/serializers.py
@@ -14,10 +14,12 @@ class CommentResponseSerializer(serializers.ModelSerializer):
 
     # Show username when author exists and comment is not anonymized; null otherwise.
     author_username = serializers.SerializerMethodField()
+    # True when the requesting user is the comment author, regardless of username visibility.
+    is_own_comment = serializers.SerializerMethodField()
 
     class Meta:
         model = Comment
-        fields = ['id', 'story_id', 'author_username', 'text', 'is_anonymized', 'created_at']
+        fields = ['id', 'story_id', 'author_username', 'text', 'is_anonymized', 'created_at', 'is_own_comment']
 
     def get_author_username(self, obj):
         if obj.is_anonymized or obj.author is None:
@@ -25,6 +27,12 @@ class CommentResponseSerializer(serializers.ModelSerializer):
         if not obj.author.is_username_public:
             return None
         return obj.author.username
+
+    def get_is_own_comment(self, obj):
+        request = self.context.get('request')
+        if request is None or not request.user.is_authenticated:
+            return False
+        return obj.author_id == request.user.id
 
 
 class LikeResponseSerializer(serializers.ModelSerializer):
--- a/backend/apps/interactions/tests/test_serializers.py
+++ b/backend/apps/interactions/tests/test_serializers.py
@@ -83,8 +83,7 @@ class TestCommentResponseSerializer:
     def test_get_is_own_comment_returns_true_for_author(self, user, story):
         comment = Comment.objects.create(story=story, author=user, text='Mine')
         request = MagicMock()
-        request.user = user
-        request.user.is_authenticated = True
+        request.user = user  # is_authenticated is a property that returns True for active users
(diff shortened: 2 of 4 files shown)


KmalM (approved):
78/78 comment-related tests pass locally. Implementation is correct and minimal:

- `obj.author_id == request.user.id` is the right comparison — uses Django's `_id` shortcut to avoid an extra DB fetch, and safely returns `False` when `author` is NULL (deleted user).
- `self.context.get('request')` guard correctly handles serializers called without request context (e.g. in service-level tests).
- `context={'request': request}` is now passed consistently in both the GET (list) and POST (create) paths.
- The private-username regression test confirms `is_own_comment` is independent of `author_username` visibility.

No blocking issues.

PR #549, opened 2026-05-06
fix: Nominatim geocoding proxy 503 — rate lock, error logging, placeholder email

### LINKED ISSUE #547: [Bug][Backend] Email verification code is not delivered after registration
## Description
After a user registers, the backend creates the user and returns a successful registration response, but the verification email may not actually be delivered.

In development, `EMAIL_BACKEND` defaults to Django console email backend, so verification emails are printed to backend logs instead of being sent to the user inbox. In production, SMTP settings depend on environment variables, but missing or invalid values such as `EMAIL_HOST_PASSWORD` or an unverified `DEFAULT_FROM_EMAIL` can cause delivery to fail.

Currently, email delivery failures are caught and only logged in `register_user`, while the API still returns success. Because of this, mobile/frontend clients move the user to the verification screen even though the user may never receive a code.

## What needs to be fixed
- Review backend email configuration for development and production environments
- Ensure production has valid SMTP/Resend credentials
- Ensure `DEFAULT_FROM_EMAIL` uses a verified sender/domain
- Make email delivery failure visible enough for debugging/monitoring
- Consider whether registration should still return success if verification email delivery fails
- Document where verification codes appear in development mode

## Why this matters
Users cannot complete registration if they never receive the verification code. Since the API still returns success, the client cannot distinguish between email sent and email failed silently, which creates a broken registration flow.

## Acceptance Criteria
- [ ] Verification emails are successfully delivered in the intended production/staging environment
- [ ] Development behavior is documented clearly, including that console backend prints codes to logs
- [ ] Missing/invalid SMTP configuration is easy to detect from logs or startup checks
- [ ] `DEFAULT_FROM_EMAIL` is configured with a verified sender/domain
- [ ] Register/resend verification flows log email delivery failures with actionable details
- [ ] Mobile/frontend clients no longer get stuck in a verification flow due to silent backend email failures

## Meta (optional)
- Effort: S/M
- Priority: High
- Area: Backend / Auth / Email Verification

### LINKED ISSUE #555: [Bug][Backend] Geocoding proxy returns 503 due to placeholder contact email banned by Nominatim
### Description

  The /geocode/ and /geocode/suggestions/ endpoints return HTTP 503 (NominatimUnavailable) for every query when NOMINATIM_CONTACT_EMAIL is unset. The default value in backend/config/settings/base.py:138 is admin@example.com, which Nominatim's bot filter hard-blocks — it returns 403 Access denied with a link to its usage policy. _nominatim_get in backend/apps/geocoding/services.py calls raise_for_status() and re-raises the 403 as our generic 503, so the failure surfaces in the frontend as a silent empty dropdown and falls back to plain text search.

 ### Reproduction:
  curl -A "LocalHistoryStoryMap/1.0 (Django proxy; contact: admin@example.com)" \
    "https://nominatim.openstreetmap.org/search?q=Paris&format=jsonv2&limit=1"                                                                         
  → HTTP 403, "Access denied. See https://operations.osmfoundation.org/policies/nominatim/"                                                          
                                                                                                                                                       
  Replacing the email with a real address returns HTTP 200.                                                                                            
                                                                                                                                                       
### Acceptance Criteria

  - [ ] NOMINATIM_CONTACT_EMAIL no longer defaults to admin@example.com; the app logs a loud warning if a real contact email is not configured
  - [ ] When _nominatim_get raises NominatimUnavailable, the upstream status code and response body are logged so future failures are diagnosable instead  
  of opaque 503s                                                                                                                                       
  - [ ] The requests.get call in _nominatim_get is moved inside the with _rate_lock: block so the 1 req/s limit is actually enforced (currently the HTTP call happens after the lock is released — services.py:35-53)                                                                                         
  - [ ] backend/.env.example documents NOMINATIM_CONTACT_EMAIL as required for local development

### Meta (optional)

_No response_

KmalM:
# 📝 Description
Fixes #547
Fixes #555

Four related changes to make email delivery and geocoding work reliably in production:

**1 — Resend HTTP API backend (replaces SMTP)**
Cloud hosts (DigitalOcean, AWS, etc.) block outbound SMTP ports 465/587. The new `ResendEmailBackend` in `common/email_backend.py` posts to `https://api.resend.com/emails` over HTTPS (port 443, never blocked). `email_service.py` is untouched — callers still use `send_mail()`. Configured via `EMAIL_BACKEND=common.email_backend.ResendEmailBackend` and `RESEND_API_KEY=re_...` in `.env.prod`.

**2 — Email delivery status surfaced to clients**
`register_user()` stays stable (returns `User` only). A new `send_registration_verification()` helper handles delivery. The `POST /auth/register/` response includes `"email_sent": true/false`. When delivery fails, the response body directs clients to `/auth/resend-verification/`.

**3 — Mailpit local email catcher**
Adds `axllent/mailpit` to `docker-compose.yml`. All dev SMTP routes through Mailpit. Developers see every verification and password-reset email at **http://localhost:8025** — no credentials required.

**4 — Nominatim proxy hardening**
Fixes the 503 loop caused by the placeholder `NOMINATIM_CONTACT_EMAIL` (hard-blocked by Nominatim). Rate lock now wraps the HTTP call (not just the check). Upstream errors are logged with status code and body. `GeocodingConfig.ready()` warns at startup if a placeholder domain is still configured.

# 📊 Type of Change
- [x] Bug fix
- [x] New feature

# 📸 Screenshots / Recordings
After `docker compose up`, open http://localhost:8025 — emails land there as they are triggered.

# ⏱️ Estimated Review Time
- [x] > 15 min

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -33,3 +33,8 @@ DEFAULT_FROM_EMAIL=Local History Story Map <noreply@yourdomain.com>
 
 # Base URL of the frontend — used to build password reset links in emails.
 FRONTEND_URL=http://localhost:3000
+
+# ── Nominatim geocoding proxy ─────────────────────────────────────────────────
+# Nominatim policy requires a real contact address in the User-Agent header.
+# Set this to the project maintainer's email for staging/production deployments.
+NOMINATIM_CONTACT_EMAIL=admin@example.com
--- a/.env.example
+++ b/.env.example
@@ -8,27 +8,24 @@ DB_HOST=127.0.0.1
 DB_PORT=3306
 MEDIA_ROOT=/app/media
 
-# ── Email (Resend) ────────────────────────────────────────────────────────────
+# ── Email (Resend HTTP API) ───────────────────────────────────────────────────
 # Provider: https://resend.com — create a free account, verify your domain,
-# and generate an API key. The username is always the literal string "resend".
+# and generate an API key.
+#
+# Production uses ResendEmailBackend (HTTPS port 443) — not SMTP — because
+# cloud hosts (DigitalOcean, AWS, etc.) block outbound SMTP ports 465/587.
 #
 # For local development, EMAIL_BACKEND is overridden to console in development.py
-# so these values are only required for staging and production.
+# so RESEND_API_KEY is not required locally.
 #
-# Steps to configure:
+# Steps to configure for production:
 #   1. Sign up at resend.com and add/verify your sending domain
 #   2. Create an API key (Resend dashboard → API Keys → Create)
-#   3. Set EMAIL_HOST_PASSWORD to the API key (starts with re_...)
+#   3. Set RESEND_API_KEY to the key (starts with re_...)
 #   4. Set DEFAULT_FROM_EMAIL to an address on your verified domain
 #   5. Set FRONTEND_URL to the deployed frontend base URL
-#
-# Resend SMTP relay — port 465 (SSL) is recommended.
-EMAIL_HOST=smtp.resend.com
-EMAIL_PORT=465
-EMAIL_HOST_USER=resend
-EMAIL_HOST_PASSWORD=re_your_api_key_here
-EMAIL_USE_TLS=False
-EMAIL_USE_SSL=True
+EMAIL_BACKEND=common.email_backend.ResendEmailBackend
+RESEND_API_KEY=re_your_api_key_here
 DEFAULT_FROM_EMAIL=Local History Story Map <noreply@yourdomain.com>
 
 # Base URL of the frontend — used to build password reset links in emails.
--- a/.env.example
+++ b/.env.example
@@ -35,6 +35,7 @@ DEFAULT_FROM_EMAIL=Local History Story Map <noreply@yourdomain.com>
 FRONTEND_URL=http://localhost:3000
 
 # ── Nominatim geocoding proxy ─────────────────────────────────────────────────
-# Nominatim policy requires a real contact address in the User-Agent header.
-# Set this to the project maintainer's email for staging/production deployments.
-NOMINATIM_CONTACT_EMAIL=admin@example.com
+# REQUIRED for local development: Nominatim's usage policy requires a real
+# contact address in the User-Agent header. Placeholder/example addresses are
+# hard-blocked with HTTP 403. Set this to your own email address.
+NOMINATIM_CONTACT_EMAIL=kemal.mahmutogullari@gmail.com
--- a/.env.prod.example
+++ b/.env.prod.example
@@ -15,7 +15,16 @@ VITE_API_URL=https://yourdomain.com/api
 NGINX_HOST=yourdomain.com
 # Optional: set to your Sentry DSN to enable error tracking.
 SENTRY_DSN=
-EMAIL_HOST=smtp.example.com
-EMAIL_PORT=587
-EMAIL_HOST_USER=
-EMAIL_HOST_PASSWORD=
+# Email delivery via Resend (https://resend.com)
+# 1. Create a Resend account and verify your sending domain.
+# 2. Generate an API key with sending permissions.
+# 3. Set EMAIL_HOST_PASSWORD to the API key (starts with re_...)
+# 4. Set DEFAULT_FROM_EMAIL to an address on your verified domain.
+EMAIL_HOST=smtp.resend.com
+EMAIL_PORT=465
+EMAIL_HOST_USER=resend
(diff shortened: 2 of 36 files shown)


AysuKeskin (changes requested):
## What Works Well

- Mailpit in `docker-compose` is a clean DX improvement. Developers get a working email flow with zero credential setup.
- Startup misconfiguration warnings in `apps.py` are a good defensive addition for production deployments.
- `transaction.atomic()` in `resend_verification()` is a correct improvement around invalidating old codes and creating the new one.
- `.gitattributes` for LF normalization is good housekeeping for cross-platform teams.
- Test coverage follows project conventions and is well-structured.

---

## Main Issue: `register_user()` Should Not Return a Tuple

Changing `register_user()` from returning `User` to returning `(user, email_sent)` is the wrong layer for this concern.

`register_user()` is a service method. Its core responsibility should be creating the user and issuing the verification code. Whether the verification email was delivered is a response concern that only `RegisterView` currently needs in order to shape the HTTP response.

This also creates an unnecessary breaking change. Any current or future caller that expects a `User` now has to unpack a tuple:

```python
# Previously valid:
user = register_user(validated_data)

# Now required:
user, _ = register_user(validated_data)
```

That makes the API easier to misuse. A caller can also silently discard the new delivery status with `user, _ = ...`, which undermines the stated goal of surfacing email delivery failures.

The better fix is to keep `register_user()` stable and move delivery-status handling into the view layer. For example, the service can return the created user and the verification code object/value explicitly, or expose a separate helper for creating the verification code, while `RegisterView` handles `send_verification_email()` and decides how to represent failure in the response.

```python
# services.py
def register_user(validated_data: dict) -> User:
    """Create an inactive user and verification code. Does not shape HTTP delivery status."""
    with transaction.atomic():
        user = User.objects.create_user(...)
        code = EmailVerificationCode.objects.create(
            user=user,
            code=EmailVerificationCode.generate_code(),
        )
    return user
```

Then `RegisterView` can send the email, catch delivery failures, log them, and return `email_sent` in the HTTP response.

This keeps the service contract stable and keeps HTTP-response-specific behavior in the HTTP layer.

---

## Secondary Issues

`resend_verification()` is only partially addressed. It now logs delivery failures with useful context, which is good, but the endpoint still always returns the same response regardless of whether email delivery succeeded. If this PR’s goal is to surface delivery failures, resend should make an explicit product/security decision here. Because the endpoint is anti-enumeration-sensitive, any `email_sent` response must avoid revealing whether the submitted email exists or is pending verification.

There is also a mismatch in the misconfiguration warning logic. `DEFAULT_FROM_EMAIL` now defaults to `noreply@yourdomain.example`, but `_warn_if_email_misconfigured()` only checks for `example.com`. That means the new placeholder sender can slip through production startup without warning. The warning should also catch `.example` / `yourdomain.example`, and there should be a test for the current default.

---

## Summary

The infrastructure and observability changes are worth keeping. The main change I would not accept as-is is the `register_user()` return type change. Keep the service contract stable, move delivery-status handling to `RegisterView`, and tighten the sender-placeholder warning.

KmalM:
All three items addressed.

**1. `register_user()` return type**

`register_user()` now returns `User` only. Email sending is extracted into a new `send_registration_verification(user) -> bool` service function that looks up the latest unused verification code and calls `send_verification_email`. `RegisterView.post()` calls both in sequence and owns the `email_sent` flag and response shaping. The service contract is stable.

**2. `resend_verification()` anti-enumeration decision**

No behavior change — the uniform-response approach is the correct call here. Added an explicit docstring explaining why `email_sent` is intentionally not surfaced: returning it would let callers distinguish "email exists and is unverified" from "email unknown or already verified", which breaks the enumeration protection.

**3. Warning check mismatch**

Updated `_warn_if_email_misconfigured` to check for both `'example.com'` and `'.example'` so the current `base.py` default (`noreply@yourdomain.example`) also triggers the warning. Added a regression test that pins the new default value specifically.

AysuKeskin (commented):
backend/config/settings/base.py:141

I think `NOMINATIM_CONTACT_EMAIL` should be required instead of defaulting to a personal email address.

```python
# before
NOMINATIM_CONTACT_EMAIL = env(
    'NOMINATIM_CONTACT_EMAIL',
    default='kemal.mahmutogullari@gmail.com'
)

# after
NOMINATIM_CONTACT_EMAIL = env('NOMINATIM_CONTACT_EMAIL')

AysuKeskin (approved):
Thanks for addressing the review comments.

The latest changes resolve the main blockers I mentioned earlier and solves the bugs mentioned in the issues. Test coverage is good as well.

Approving this PR.

PR #562, opened 2026-05-07
fix:map-pins-mobile

### LINKED ISSUE #552: [Bug][Mobile] Map pins overlap when multiple items share the same location
# Bug Report

## Description
When two or more map items are created at the exact same location, the mobile map UI renders their pins directly on top of each other. Only the top marker is practically tappable/visible, so users cannot reliably discover or open all items attached to that coordinate.

The mobile map should handle duplicate coordinates by showing that multiple pins exist at the same point and allowing the user to choose among them. A clustered marker, stacked/offset pins, or a bottom sheet/list after tapping the shared location would all be acceptable as long as every item remains reachable.

## Acceptance Criteria
- [ ] When multiple mobile map items have the same latitude/longitude, the map indicates that more than one item exists at that point.
- [ ] Tapping the shared-location marker gives access to each item at that coordinate.
- [ ] Existing single-pin behavior remains unchanged for locations with only one item.
- [ ] The solution works at common zoom levels without pins hiding each other or becoming untappable.
- [ ] Add a regression test or documented manual test case for duplicate-coordinate pins, if automated coverage is not practical.

## Meta (optional)
Effort: S
Priority: P2
Area: Mobile map UI

MertErenKaplan:
# 📝 Description
Fixes #552

Fixed overlapping mobile map pins when multiple stories share the exact same latitude and longitude.

This PR:
- Groups duplicate-coordinate story pins into a single multi-story marker
- Shows the marker count for shared-location pins
- Keeps each story accessible through the existing cluster preview list
- Preserves the existing single-story marker behavior for unique coordinates
- Improves marker tap scrolling so the screen lands on the story preview panel more reliably
- Adds regression coverage for duplicate-coordinate map pins

# 📊 Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Refactoring
- [ ] Documentation
 
# 📸 Screenshots / Recordings
N/A

Manual test:
Verified that duplicate-coordinate stories are grouped into a count marker on the mobile map, tapping it shows all stories at that location, and each story remains openable. Also verified that unique-coordinate pins keep the existing single-story behavior and marker taps scroll to the story preview panel.

# ⏱️ Estimated Review Time
- [ ] < 5 min
- [x] 5-15 min
- [ ] > 15 min

### DIFF
--- a/mobile/src/features/map/data/repositories/__tests__/MapRepository.test.ts
+++ b/mobile/src/features/map/data/repositories/__tests__/MapRepository.test.ts
@@ -8,6 +8,10 @@ jest.mock('../../../../stories/application/services', () => ({
 }));
 
 describe('MapRepositoryImpl', () => {
+  beforeEach(() => {
+    jest.mocked(storyService.getMapPins).mockReset();
+  });
+
   it('returns one marker per story pin without clustering', async () => {
     const getMapPins = jest.mocked(storyService.getMapPins);
     getMapPins.mockResolvedValue([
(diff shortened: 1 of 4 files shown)


OguzSemih (approved):
The issue regarding overlapping mobile map pins for stories sharing identical coordinates has been resolved successfully. The updated implementation groups duplicate-coordinate stories into a unified marker structure while preserving access to each individual story through the existing preview interaction flow.

The distinction between grouped markers and single-story markers is handled consistently, and the revised marker interaction behavior improves overall usability on the mobile map interface. Manual testing confirms that shared-location stories remain fully accessible and that unique-coordinate pins continue to function as expected.

The implementation addresses both the visual overlap problem and the related interaction concerns without disrupting the existing map experience. Test coverage and regression validation further support the reliability of the solution. Ready to merge.

PR #573, opened 2026-05-07
feat(frontend): add map auto-zoom and pin clustering

omarreis01:
# Description
Fixes #458

Adds auto-zoom (fit bounds) and pin clustering to the story map.

- **Auto-zoom**: when ≥ 1 stories load, the map fits to the bounds of all visible markers. Single-marker case caps `maxZoom: 15` to avoid street-level zoom; multi-marker case applies `padding: [40, 40]`. Empty case keeps the Istanbul default. Re-fits when the feature set changes.
- **Clustering**: replaces the bare GeoJSON layer with `leaflet.markercluster` integrated imperatively via `useMap()` — the React wrapper's peer-dep with `react-leaflet 5` is unreliable, so the library is wired directly.
- **Lazy popup rendering**: `bindPopup(() => featurePopupHtml(feature))` so popup HTML is only computed when a marker is actually opened, not for every marker on every feature change.
- Removes dead `pointToLayer` / `onEachFeature` helpers (no longer used by production code) and the matching dead test branches.
- Hoists `leaflet.markercluster*` mocks to the shared `src/test/setup.js` so they're declared once.

New dependency: `leaflet.markercluster ^1.5.3`.

# Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Documentation

# Acceptance Criteria / Checklist
- [x] Project builds successfully
- [x] My code follows the style guidelines of this project
- [x] I have commented my code, especially in hard-to-understand parts
- [x] I have performed a self-review of my code
- [x] I have added/updated tests
- [x] New and existing unit tests pass locally with my changes

# Screenshots / Recordings

# Estimated Review Time
- [ ] < 5 min
- [x] 5-15 min
- [ ] > 15 min

Issue #389, opened 2026-04-16
[Feature][Frontend] Implement Geolocation-Based Radius Filtering in Filter Tab


sezindogan:
### Description

To enhance the discovery of local content, we need to implement a geolocation-based filtering system within the search/filter tab. This feature will allow users to filter stories based on their current physical proximity. The interface should offer predefined radius options (e.g., 1 km, 10 km, 100 km) to narrow down the Story results relative to the user's current coordinates.

This feature will leverage the existing map-based capabilities of the platform and requires integration with the browser's Geolocation API to determine the user's starting point for the radius calculation.

### Implementation Details

- UI Component: Add a "Proximity" or "Distance" section to the Filter Tab.
- Input Type: Use a radio button group or a dropdown menu for the predefined radius options.
- Coordinate Retrieval: Request user permission to access current location via navigator.geolocation.
- Standard Compatibility: Ensure that location data handles coordinate precision appropriately to maintain consistency with the map-based search bar and autocomplete features

### Acceptance Criteria

- [ ] The Filter Tab includes a new section for Radius Filtering with the following options: 1 km, 10 km, 100 km, and "Anywhere" (Off).

- [ ] Upon selecting a radius, the application prompts the user for location permissions if not already granted.

- [ ] If the user denies location access, an appropriate error message or "Location Disabled" state is displayed within the filter section.

- [ ] Selecting a radius option triggers a filtered API request that passes the user's current latitude/longitude and the chosen distance parameter.

- [ ] The filter state is cleared or reset when the user selects "Anywhere" or manually resets the filter tab.

- [ ] The UI provides visual feedback (e.g., a loading spinner) while the geolocation is being fetched or the filtered results are loading.

- [ ] Unit tests are written.

PR #575, opened 2026-05-08
feat(frontend): fit map to geocoded bbox with priority over markers

### LINKED ISSUE #574: [Feature][Frontend] Map auto-zoom should follow geocoded location filter
## Description

The web map's auto-zoom (added in PR #573) currently fits only to **marker bounds**. When the user picks a location from the geocoding autocomplete in the filter panel, the resulting bbox is correctly forwarded to the backend (and stories are filtered by geography), but `<FitBoundsToFeatures>` in `frontend/src/components/MapView/MapView.jsx` ignores the bbox and only sees the returned markers.

Concrete consequences:
- If the geocoded area has 0 matching stories (e.g., a remote address with no submissions), the map does **nothing** — the user has no visual confirmation that their location filter even applied.
- If the geocoded area has 1–2 matching stories far apart, the map zooms to those markers' extent rather than the area the user actually searched for.

Mobile already handles this via `getRegionForLocationBounds` in `MapScreen.tsx`, which prefers `filters.locationBounds` over marker bounds. This issue closes the parity gap on web.

## Scope

- Add an optional `bbox` prop to `<MapView>` and `<FitBoundsToFeatures>`.
- Wire it from `MapPage.jsx` (`useFilterState()` already exposes `latMin/latMax/lngMin/lngMax`).
- In `<FitBoundsToFeatures>`, prefer the bbox tier when present; fall back to marker-bounds tier; keep the existing default Istanbul view when neither applies.
- Preserve the existing `lastFitKeyRef` no-op guard so refetches with the same context do not re-fit.

## Acceptance Criteria

- [ ] `MapPage.jsx` passes `bbox = { latMin, latMax, lngMin, lngMax }` (or `null`) to `<MapView>` based on URL params.
- [ ] When `bbox` is provided, the map fits to that bbox **even if 0 markers are returned** — the user sees the area they searched for.
- [ ] When `bbox` is not provided, the existing marker-bounds auto-zoom behavior is unchanged (no regression vs PR #573).
- [ ] When neither `bbox` nor markers are present, the map keeps the default Istanbul view (no regression).
- [ ] The `lastFitKeyRef` no-op guard is extended to include the bbox in the key, so the same bbox + same marker set does not re-fit.
- [ ] Vitest coverage for: bbox-only fit (zero markers), bbox + markers fit, marker-only fit, empty-state no-op, and no-op guard with unchanged bbox.

## Context / Related

- PR #573 — added marker-bounds auto-zoom + clustering (closes #458).
- PR #531 — added the Nominatim geocoding hook + URL bbox params (already merged).
- Mobile reference: `mobile/src/features/map/presentation/screens/MapScreen.tsx` (`getRegionForLocationBounds`, `getAutoZoomRegion`).
- Mobile twin issue: #491 (closed).

omarreis01:
# Description
Fixes #574

When a user picks a location from the geocoding autocomplete, the resulting bbox is forwarded to the backend (and stories are filtered geographically), but the web map's auto-zoom previously only saw returned markers. So a remote area with 0 matches gave no visual confirmation, and an area with 1–2 distant matches zoomed to the markers instead of the searched region.

This PR adds bbox-priority auto-zoom on the web, matching the mobile `getRegionForLocationBounds` behavior:

- `MapPage` derives `bbox = { latMin, latMax, lngMin, lngMax }` (or `null`) from existing URL params via `useMemo` and passes it to `<MapView>`.
- `<FitBoundsToFeatures>` now prefers a valid bbox over marker bounds. When neither is present, the default Istanbul view is preserved.
- The `lastFitKeyRef` no-op guard keys on bbox + feature ids together, so identical refetches don't re-fit while a changing bbox does.

# Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Documentation

# Screenshots / Recordings

# Estimated Review Time
- [x] < 5 min
- [ ] 5-15 min
- [ ] > 15 min

### DIFF
--- a/frontend/src/components/MapView/MapView.jsx
+++ b/frontend/src/components/MapView/MapView.jsx
@@ -57,22 +57,47 @@ function featureLatLng(feature) {
   return [lat, lng];
 }
 
-function FitBoundsToFeatures({ features }) {
+function isValidBbox(bbox) {
+  return (
+    bbox != null &&
+    Number.isFinite(bbox.latMin) &&
+    Number.isFinite(bbox.latMax) &&
+    Number.isFinite(bbox.lngMin) &&
+    Number.isFinite(bbox.lngMax)
+  );
+}
+
+function FitBoundsToFeatures({ features, bbox }) {
   const map = useMap();
-  // Skip re-fitting when the feature set is unchanged — refetches that
-  // return the same stories shouldn't snap the map back over the user's pan.
+  // Skip re-fitting when the inputs are unchanged — refetches that return the
+  // same bbox + stories shouldn't snap the map back over the user's pan.
   const lastFitKeyRef = useRef(null);
 
   useEffect(() => {
     if (!map) return;
-    if (features.length === 0) {
+    const hasBbox = isValidBbox(bbox);
+    if (!hasBbox && features.length === 0) {
       lastFitKeyRef.current = null;
       return;
     }
-    const fitKey = features.map((f) => f.id).join("|");
+    const bboxKey = hasBbox
+      ? `${bbox.latMin}|${bbox.latMax}|${bbox.lngMin}|${bbox.lngMax}`
+      : "";
+    const featuresKey = features.map((f) => f.id).join("|");
+    const fitKey = `${bboxKey}::${featuresKey}`;
     if (fitKey === lastFitKeyRef.current) return;
     lastFitKeyRef.current = fitKey;
 
+    // Bbox takes priority so the user always sees the area they searched
+    // for, even if no markers fall inside it.
+    if (hasBbox) {
+      const bounds = L.latLngBounds([
+        [bbox.latMin, bbox.lngMin],
+        [bbox.latMax, bbox.lngMax],
+      ]);
+      map.fitBounds(bounds, { padding: FIT_BOUNDS_PADDING });
+      return;
+    }
     const latlngs = features.map(featureLatLng).filter(Boolean);
     if (latlngs.length === 0) return;
     const bounds = L.latLngBounds(latlngs);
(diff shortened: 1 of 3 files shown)


sezindogan (approved):
## Overview
                                                                                                                                                       
  Single commit on top of PR #573. Adds a bbox prop to MapView so that when the user applies a geocoded location filter (with latMin/latMax/lngMin/lngMax from useFilterState), the map fits to the searched area instead of (or in priority over) the marker bounds. MapPage derives bbox from filter state via useMemo and passes it through. Auto-zoom de-dup key now includes the bbox so a bbox change re-fits even with the same marker set.

## Strengths

  - Necessary follow-up to #573. PR #573 fit-to-markers only. Before this PR, searching a location that returned zero stories — or stories far from the searched area — left the map sitting on its previous view, which is a real UX bug. Fitting to the geocoded bbox is the right fix and the natural completion of #573, not scope creep.                                                                                                                 
  - Mobile parity on the bbox path. Mobile (MapScreen.tsx:461 getAutoZoomRegion) already prioritizes locationBounds over marker bounds in exactly the same way — same precedence, same intent. This PR pulls web onto the model mobile already shipped.                                                    
  - Implementation is small and well-contained: isValidBbox guard, composite fitKey (bboxKey::featuresKey) preserves the "don't snap back over user pan" property from #573, partial-bbox falls through to marker bounds.                                                                                
  - Test coverage is thorough — bbox-only, bbox + markers (priority), no bbox (fallback), neither, no re-fit on identical re-render, re-fit on bbox change, and the partial-bbox fallthrough. All the meaningful branches.                                                                               
                            
 ## Issues                                                                                                                                               
                            
  - No blocking issues for this PR. The remaining auto-zoom parity gap with mobile — proximity filter (latitude/longitude/radiusKm), which mobile handles via getRegionForProximityFilter — is out of scope here and will be picked up alongside the proximity filter work tracked in issue #389 . Calling it out so the next author wires the auto-zoom branch in at the same time and we don't drift from mobile again.                               
                            
 ## Minor (fine to leave as-is)

  - MapPage already returns null when any bbox field is missing, so MapView's partial-bbox guard is defensive only — fine, the test pins the contract. 
  - Bbox prop is a fresh object each render; the value-based fitKey correctly absorbs that, so referential instability isn't a problem.

This branch is ready to be merged.

PR #577, opened 2026-05-08
Feat/frontend proximity filtering

sezindogan:
# 📝 Description
Fixes #389                                                                                                                                            
                                                                                                                                                       
  Adds a Distance section to the filter panel that lets users narrow stories to a radius around their current location (Anywhere / 500 m / 1 km / 10 km / 100 km — matching mobile). Selecting a radius prompts the browser for geolocation permission; on success the coordinates and chosen radius_km flow through to /stories/feed/, /stories/search/, and /stories/map/. The map auto-zooms to the proximity circle when no geocoded bbox is set, completing the parity gap left after #573 / #574 between web and mobile (getRegionForProximityFilter on mobile).

 Uses navigator.geolocation only — no new Nominatim/OSM calls. The existing backend Nominatim proxy is untouched.                                     
  
##  New features                                                                                                                                         
                                                            
  - Distance / proximity filter with five predefined options matching mobile (Anywhere, 500 m, 1 km, 10 km, 100 km)                                    
  - Browser geolocation permission flow with loading spinner, "Location disabled" error on denial, and "Current location unavailable" fallback for timeouts / missing API                                                                                                                               
  - Denied selection stays highlighted so subsequent clicks retry the permission flow instead of snapping back to Anywhere; Apply gates on resolved coordinates so a half-applied filter never reaches the backend                                                                                       
  - Removable "Within X" chip in the active-filters strip (formatted as Within 500 m for sub-km radii); counted in the filter badge
  - Map auto-zoom to a circle around the user when proximity is active and no bbox is set; precedence is bbox > proximity > markers, with bounds math (latDelta = radiusKm/111, lngDelta scaled by cos(lat), 1.2× padding) mirroring mobile                                                                
                                                                                                                                                       
 ## Modified files                                                                                                                                       
                                                            
  - services/deviceLocationService.js (new) — wraps navigator.geolocation.getCurrentPosition with a granted | denied | unavailable envelope mirroring mobile/src/core/services/deviceLocation.ts                
  - hooks/useFilterState.js — reads/writes latitude, longitude, radius_km URL params; treats proximity as an active filter only when all three are present                                                                                                                                              
  - services/storyService.js — forwards proximity params on feed, search, and map endpoints (omits when partial)
  - components/SearchFilter/FilterPanel.jsx — new Distance radio group, draft state, permission flow, retry-on-reclick behavior                        
  - components/SearchFilter/SearchFilter.jsx — plumbs proximity through; treats it as a removable filter via the proximity chip key                    
[...]

PR #579, opened 2026-05-08
feat(frontend): add location search and autocomplete to story submission map

### LINKED ISSUE #360: [Feature][Frontend] Implement Location Search and Autocomplete in Story Submission Map
### Description

Currently, in the story submission form, users are required to manually pan and zoom across the map to drop a pin on their desired location. While this works for general areas, it creates friction when users want to tag a specific address or point of interest.

To improve the user experience, we need to introduce a map-based search bar with autocomplete capabilities. This will allow users to text-search a location, select the correct address from a dropdown, and have the map automatically pan, zoom, and drop a pin at the target coordinates. The existing manual map navigation and pin-dropping functionality must remain fully operational.

### Acceptance Criteria

- [ ] As the user types, the search bar must display a list of suggested addresses or points of interest, appearing after at least 3 characters are entered.
- [ ] Selecting an address from the dropdown must trigger the map to automatically pan and zoom to the chosen location.
- [ ] A location pin must be automatically dropped at the precise coordinates of the selected search result upon selection.
- [ ] The existing manual pan, zoom, and click-to-drop pin functionality must remain fully operational and the last chosen pin must persist, whether it is manual or search based.
- [ ] The story submission form's latitude and longitude fields must automatically update to match the coordinates of the selected or manually dropped pin.
- [ ] Unit tests must verify that API calls are debounced (e.g., 300ms) to prevent excessive requests during active typing.
- [ ] Unit tests must confirm that selecting a suggestion correctly updates the application state with the expected coordinate object.
- [ ] The system must handle "no results found" states gracefully and include unit tests for API error responses to ensure the map remains interactive if the search service fails.

### Meta (optional)

_No response_

omarreis01:
# Description
Fixes #360

The story submission map previously required users to manually pan and zoom to drop a pin — frustrating when you have a specific address in mind. This PR adds a Nominatim-backed search bar above the map. Typing 3+ characters triggers a debounced suggestions list; selecting a result drops a pin at the bbox centre, updates the form's `{lat, lng}` state, and pans/zooms the map to that location. Manual click-to-drop is preserved untouched — both paths write to the same controlled `value`, so the last chosen pin persists whichever way it was placed.

Reuses existing infrastructure:
- `useLocationSuggestions` hook handles the 300 ms debounce, 3-char minimum, and error-swallowing.
- `geocodingService.searchLocationSuggestions` is the backend Nominatim proxy and already caches by query in memory.
- No backend changes required.

A small `MapController` child component wraps `useMap().setView` and only fires when a *new* target is set (i.e. on suggestion select), so manual clicks never snap the map back.

# Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Documentation

# Screenshots / Recordings

# Estimated Review Time
- [ ] < 5 min
- [x] 5-15 min
- [ ] > 15 min

---

### Notes for reviewer

- **Known duplication, intentionally not addressed here:** the listbox/keyboard-nav/click-outside markup mirrors what already exists in `SearchFilter/FilterPanel.jsx`. Extracting a shared `<LocationSuggestionsCombobox>` was scoped out — `FilterPanel.jsx` is being modified by #577 right now, so a refactor would conflict. Worth a follow-up issue once #577 lands.
- **Pin coordinate** is the centre of the suggestion's bbox. The backend Nominatim proxy at `/geocode/suggestions/` doesn't expose the raw `lat`/`lon` fields (only `bbox`), so centre-of-bbox is the precise location for POI-level results and a reasonable approximation for area-level results.

### DIFF
--- a/frontend/src/components/MapPicker/MapPicker.jsx
+++ b/frontend/src/components/MapPicker/MapPicker.jsx
@@ -7,6 +7,7 @@ import { Loader2 } from "lucide-react";
 import { Input } from "@/components/ui/input";
 import { useLocationSuggestions } from "@/hooks/useLocationSuggestions";
 import { cn } from "@/lib/utils";
+import { bboxCenter } from "./bbox";
 import markerIcon2x from "leaflet/dist/images/marker-icon-2x.png";
 import markerIcon from "leaflet/dist/images/marker-icon.png";
 import markerShadow from "leaflet/dist/images/marker-shadow.png";
@@ -21,13 +22,6 @@ L.Icon.Default.mergeOptions({
 const ISTANBUL_CENTER = [41.0082, 28.9784];
 const DEFAULT_ZOOM = 12;
 const SELECTED_ZOOM = 15;
-
-function bboxCenter(bbox) {
-  return {
-    lat: (bbox.latMin + bbox.latMax) / 2,
-    lng: (bbox.lngMin + bbox.lngMax) / 2,
-  };
-}
 
 function ClickHandler({ onChange }) {
   useMapEvents({
@@ -58,24 +52,27 @@ function MapPicker({ value, onChange }) {
   const [target, setTarget] = useState(null);
   const containerRef = useRef(null);
 
-  const { suggestions, isLoading, clearSuggestions } = useLocationSuggestions(query);
+  const { suggestions, isLoading, hasSearched, cancel } = useLocationSuggestions(query);
 
   // Dismiss the suggestions panel when clicking outside the search field.
   useEffect(() => {
     function handleClickOutside(e) {
       if (containerRef.current && !containerRef.current.contains(e.target)) {
-        clearSuggestions();
+        cancel();
         setActiveIndex(-1);
       }
     }
     document.addEventListener("mousedown", handleClickOutside);
     return () => document.removeEventListener("mousedown", handleClickOutside);
-  }, [clearSuggestions]);
+  }, [cancel]);
 
   function selectSuggestion(s) {
+    if (!s.bbox) return;
     const center = bboxCenter(s.bbox);
+    // Suppress the next search for this exact query so writing the title back
+    // into the input doesn't re-fetch and reopen the dropdown 300 ms later.
+    cancel(s.title);
     setQuery(s.title);
-    clearSuggestions();
     setActiveIndex(-1);
     onChange(center);
     setTarget({ ...center, zoom: SELECTED_ZOOM });
@@ -94,14 +91,14 @@ function MapPicker({ value, onChange }) {
       selectSuggestion(suggestions[activeIndex]);
     } else if (e.key === "Escape") {
       e.preventDefault();
-      clearSuggestions();
+      cancel();
       setActiveIndex(-1);
     }
   }
 
   const trimmed = query.trim();
   const showNoResults =
-    trimmed.length >= 3 && !isLoading && suggestions.length === 0;
+    trimmed.length >= 3 && !isLoading && hasSearched && suggestions.length === 0;
 
   return (
     <>
@@ -117,6 +114,7 @@ function MapPicker({ value, onChange }) {
           onKeyDown={handleKeyDown}
           aria-label="Search for a location"
           aria-autocomplete="list"
+          aria-controls="mappicker-listbox"
           aria-expanded={suggestions.length > 0}
           aria-activedescendant={
             activeIndex >= 0
@@ -134,6 +132,7 @@ function MapPicker({ value, onChange }) {
         )}
         {suggestions.length > 0 && (
           <ul
+            id="mappicker-listbox"
             role="listbox"
             aria-label="Location suggestions"
             className="absolute left-0 right-0 top-full z-[1100] mt-1 max-h-56 overflow-y-auto rounded-md border bg-background shadow-md"
--- a/frontend/src/components/MapPicker/MapPicker.jsx
+++ b/frontend/src/components/MapPicker/MapPicker.jsx
@@ -1,11 +1,16 @@
-import { MapContainer, TileLayer, Marker, useMapEvents } from "react-leaflet";
+import { useEffect, useRef, useState } from "react";
+import { MapContainer, TileLayer, Marker, useMap, useMapEvents } from "react-leaflet";
 import "leaflet/dist/leaflet.css";
 import L from "leaflet";
+import { Loader2 } from "lucide-react";
+
+import { Input } from "@/components/ui/input";
+import { useLocationSuggestions } from "@/hooks/useLocationSuggestions";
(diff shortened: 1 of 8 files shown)


sezindogan:
## Overview                                                                                                                                             
                            
  Adds a debounced location search + autocomplete to MapPicker for story submission. The feature routes all geocoding through the existing Django backend proxy (which itself calls Nominatim), adds keyboard navigation and ARIA roles to the suggestions list, and animates the map to the selected bbox centre. Companion work includes a new useLocationSuggestions hook, a geocodingService with in-memory caching, and the Django geocoding app with rate-limiting and server-side caching.                                                                                                               
                            
 ## Strengths
           
  - Nominatim policy compliance is solid. No direct browser→Nominatim calls; every request goes through /geocode/ or /geocode/suggestions/. The Django proxy sets a correct User-Agent (LocalHistoryStoryMap/1.0 (Django proxy; contact: <email>)) sourced from NOMINATIM_CONTACT_EMAIL in env, honours the 1 req/s rate limit, and caches results for 24 h. OSM tile attribution is present in TileLayer.
  - Good layering. Service (geocodingService.js) → hook (useLocationSuggestions) → component is clean. Each layer has its own focused test file.       
  - Keyboard + ARIA. ArrowUp/Down, Enter, Escape all work; aria-autocomplete, aria-expanded, aria-activedescendant, role="listbox", role="option", and aria-selected are all wired correctly.                                                                                                               
  - Race-condition safety in the hook. cancelled ref + clearTimeout cleanup prevents stale results from painting the UI when queries change quickly.   
  - MapController is well isolated. The lastTargetRef guard prevents repeated setView calls on re-renders without a new target.                        
  - Test coverage is broad. Debounce timing, API error path, keyboard navigation, pin-drop vs search-select, and the "no results" state are all covered.                                                                                                                                             
                                                                                                                                                                                                                                                                              
 ## Bugs (found in manual testing)
                                
###  Bug 1 — Dropdown reopens after selection (MapPicker.jsx:65)
                                                                                                                                                       
 Root cause: selectSuggestion calls setQuery(s.title) before clearSuggestions(). Because query changes to the selected title (a valid place, ≥ 3 chars), useLocationSuggestions' useEffect sees the change, sets isLoading = true, and schedules a fresh 300 ms debounce search. clearSuggestions() does hide the list in that same render batch — but 300 ms later the search fires, returns results, and the dropdown reopens. From the user's perspective the menu never properly closes.
            
 ### Bug 2 — "No results found" persists after pin drop (MapPicker.jsx:90-98)
                                                                          
After successful location search and suggestion and manually closing the dropdown as explained in bug 1,  under the search bar 'No results found' text appears and does not go away. This might be due to manual closing the dropdwn menu or entirely different reason but regardless, that message should not be there.                                                                                                                                
                                                                                                                                               
##  Issues                    
        
 1. Null-bbox crash in selectSuggestion (MapPicker.jsx:65)
                                                                                                                                                       
  bboxCenter is called unconditionally, but s.bbox can be null — the service explicitly maps bbox: null for results without a bounding box, and the test "sets bbox to null when backend returns null bbox" verifies this. Any suggestion with a null bbox will throw a TypeError when selected.         
             
  Fix: either filter null-bbox suggestions out of the list before rendering, or fall back to lat/lon fields from the raw result.                                                                                                                                                   
  Better: exclude bbox-less suggestions upstream in searchLocationSuggestions or render them as disabled in the list.                                  
  

 2. clearSuggestions doesn't cancel the in-flight debounce (useLocationSuggestions.js:40)                                                                
  
  clearSuggestions is setSuggestions([]) only. If called mid-debounce (e.g., user presses Escape 200 ms into a 300 ms wait), the timer still fires 100 ms later, the spinner remains visible, and if results arrive they repopulate the list. This is observable when pressing Escape quickly while loading.
Fix: expose a cancel from the hook that both clears suggestions and cancels the pending timer, or accept an enabled flag.                           

##  Minor
       
  - No aria-controls on <Input> linking it to the role="listbox". Some screen readers use this relationship to associate the input with the suggestion list. Add aria-controls="mappicker-listbox" on the input and id="mappicker-listbox" on the `<ul>`.                                                     
  - Frontend in-memory cache has no TTL or size cap (geocodingService.js:3-4). Identical queries within the same page load never hit the backend, which is good, but the map grows unboundedly over a long session. Low priority given the server-side 24 h cache, but worth noting.                        
  - bboxCenter is not exported/tested directly. It's a pure function with a clear contract; a unit test would catch the null case described above.
  - Test for "dropdown closes after selection" is missing. MapPicker.test.jsx verifies that onChange is called and the listbox disappears, but there is no assertion that the listbox stays closed after the 300 ms re-search completes (because the re-search isn't mocked to return results in that scenario). Once Bug 1 is fixed, add a test that stubs searchLocationSuggestions to return results for the selected title and asserts the listbox does not reappear.

omarreis01:
Thanks for the thorough review @sezindogan! Pushed [6f2ebe0](https://github.com/bounswe/bounswe2026group4/pull/579/commits/6f2ebe0) addressing the findings:

**Bugs**
- **Bug 1** (dropdown reopens after select) — `useLocationSuggestions` now exposes `cancel(suppressQuery)`. `selectSuggestion` calls it before writing the chosen title back, so the next debounce for that exact query is skipped and the listbox stays closed.
- **Bug 2** ("No results found" lingers) — added a `hasSearched` flag; the no-results hint only shows after a real search returns empty, never right after a programmatic dismiss/select.

**Issues**
- **Null-bbox crash** — `searchLocationSuggestions` now drops bbox-less results upstream, and `selectSuggestion` has a defensive `if (!s.bbox) return` in case one slips through.
- **`cancel` aborts in-flight debounce** — `cancel()` clears the pending timer, resets `isLoading`/`hasSearched`, and bumps a version counter so any already-running fetch is treated as stale and won't write back.

**Minor**
- `aria-controls="mappicker-listbox"` on the input + `id="mappicker-listbox"` on the `<ul>`.
- `bboxCenter` extracted to `MapPicker/bbox.js` with its own unit tests (incl. the null-coord and zero-size cases).
- New MapPicker tests cover the Bug 1 / Bug 2 regressions, the null-bbox defensive guard, and the aria wiring.

**Deferred (intentionally)**
- Frontend cache TTL/cap — agreed it's low-prio given the 24 h server-side cache; will spin off as a follow-up issue.
- The `clearSuggestions` → `cancel` rename in `FilterPanel.jsx` — kept `clearSuggestions` as a back-compat alias on the hook since #577 is touching that file concurrently. Same reason the listbox/keyboard-nav extraction is still deferred.

All 910 tests pass; lint + build clean.

sezindogan (approved):
Verdict: Ready to merge   
All bugs and issues from the review are resolved. Tests are comprehensive and passing. The one open item (cache TTL) was explicitly acknowledged and is low priority. This is good to approve.

PR #583, opened 2026-05-08
feat(frontend): add account deletion with confirmation dialog

### LINKED ISSUE #181: [Feature][Frontend] Account deletion with confirmation and anonymization notice
## Description

Registered users must be able to delete their accounts (requirement 1.1.2.11). When an account is deleted, the user's stories are anonymized (contributor name replaced with "Anonymous") unless the user explicitly requests that their stories also be deleted. This is an irreversible action that requires careful UX to prevent accidental deletion (requirement 2.2.3).

### What needs to be built

1. **Account deletion option:** A "Delete Account" button in the profile settings or a dedicated settings/account page. This should be visually de-emphasized (not a prominent red button at the top) to prevent accidental clicks.

2. **Confirmation modal:** When the user clicks "Delete Account", a modal appears explaining exactly what will happen:
   - "Your account will be permanently deleted."
   - "Your stories will remain on the platform but your name will be replaced with 'Anonymous'."
   - An optional checkbox: "Also delete all my stories" (with a warning that this is irreversible)
   - Password re-entry field for security verification
   - "Delete My Account" confirmation button (red/danger styling)
   - "Cancel" button

3. **Submission:** On confirmation, calls `DELETE /users/:id` with the password and the delete-stories flag. On success, clears auth state (logout) and redirects to the home page with a toast ("Your account has been deleted.").

4. **API service:** Add `userService.deleteAccount(password, deleteStories)` to the user service layer.

### Why this matters

Account deletion is both a user right and a compliance consideration (requirement 1.1.2.11). The anonymization approach preserves community content while respecting the user's wish to leave the platform. The confirmation flow with password re-entry prevents accidental or unauthorized deletions (requirement 2.2.3).

## Acceptance Criteria

- [ ] "Delete Account" button accessible from profile/settings page
- [ ] Confirmation modal explains consequences (anonymization vs. story deletion)
- [ ] Optional checkbox to also delete all user's stories
- [ ] Password re-entry field required before deletion
- [ ] Danger-styled confirmation button + cancel button
- [ ] On confirm: calls `DELETE /users/:id`, clears auth state, redirects to home
- [ ] Success toast displayed after deletion
- [ ] Loading state during deletion request
- [ ] API service: `userService.deleteAccount(password, deleteStories)`
- [ ] Unit tests: modal rendering, checkbox toggle, password validation, deletion flow, redirect

omarreis01:
# Description
Fixes #181

Adds the ability for a registered user to permanently delete their own account (requirement 1.1.2.11), gated by a confirmation modal that requires password re-entry to prevent accidental or unauthorised deletion (requirement 2.2.3).

The trigger lives in a new "Danger zone" section at the bottom of the profile-edit form — visually de-emphasised (small outline button, not a prominent red one) so it's hard to hit by accident. Clicking it opens an AlertDialog that explains what happens: by default the user's stories stay on the platform but are anonymised (contributor → "Anonymous"); a checkbox lets the user opt into a hard delete that also wipes their stories, comments, and uploads.

After password confirmation:
- `userService.deleteAccount(password, deleteStories)` calls `DELETE /users/me/` with `{ password, hard_delete, refresh }`. The refresh token is forwarded so the backend blacklists it.
- On success: `tokenStore.clear()` drops auth state via the existing `auth:logout` window event, the dialog closes, a toast confirms deletion, and the user is redirected to `/`.
- Backend errors (e.g., wrong password) are surfaced inline so the user can correct and retry without losing the dialog state.

# Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Documentation

# Screenshots / Recordings

# Estimated Review Time
- [ ] < 5 min
- [x] 5-15 min
- [ ] > 15 min

---

### Notes for reviewer

- **Backend mapping:** `deleteStories` (UI checkbox) maps to backend `hard_delete`. Default state of the checkbox is unchecked → `hard_delete: false` → soft delete → stories anonymised. Matches the backend contract in `apps/users/services.py::delete_account`.
- **Auth-state cleanup:** uses `tokenStore.clear()` directly rather than `authService.logout()`, because the deletion request already blacklists the refresh token server-side, so an additional `POST /auth/logout/` would be redundant. The `auth:logout` window event still fires, so `AuthContext` drops the in-memory user.
- **Dialog state hygiene:** the form body is mounted lazily (`{open && <DeleteAccountForm />}`) so password / checkbox / error state are reset cleanly on every open/close cycle without an explicit `useEffect` reset.

### DIFF
--- a/frontend/src/components/Profile/DeleteAccountDialog.jsx
+++ b/frontend/src/components/Profile/DeleteAccountDialog.jsx
@@ -24,7 +24,6 @@ function DeleteAccountForm({ onOpenChange }) {
   const navigate = useNavigate();
   const { toast } = useToast();
   const [password, setPassword] = useState("");
-  const [deleteStories, setDeleteStories] = useState(false);
   const [submitting, setSubmitting] = useState(false);
   const [error, setError] = useState("");
 
@@ -37,7 +36,9 @@ function DeleteAccountForm({ onOpenChange }) {
     setSubmitting(true);
     setError("");
     try {
-      await deleteAccount(password, deleteStories);
+      // Aligned with mobile: account deletion always hard-deletes the user's
+      // stories, comments, and uploads — no opt-in checkbox.
+      await deleteAccount(password, true);
       // tokenStore.clear() dispatches an "auth:logout" event that the
       // AuthContext listens for to drop the in-memory user.
       clearTokens();
@@ -67,27 +68,8 @@ function DeleteAccountForm({ onOpenChange }) {
 
       <div className="mt-4 space-y-4">
         <p className="text-sm text-muted-foreground">
-          By default, your stories will remain on the platform but your name
-          will be replaced with <span className="font-medium">Anonymous</span>.
+          Your stories, comments, and likes will also be permanently deleted.
         </p>
-
-        <label className="flex items-start gap-2 text-sm">
-          <input
-            type="checkbox"
-            checked={deleteStories}
-            onChange={(e) => setDeleteStories(e.target.checked)}
-            disabled={submitting}
-            className="mt-0.5 h-4 w-4 rounded border-input"
-            data-testid="delete-stories-checkbox"
-          />
-          <span>
-            Also delete all my stories.
-            <span className="block text-xs text-muted-foreground">
-              Story content, comments, and uploads will be permanently removed.
-              This is irreversible.
-            </span>
-          </span>
-        </label>
 
         <div className="space-y-1.5">
           <Label htmlFor="delete-account-password">
@@ -103,11 +85,17 @@ function DeleteAccountForm({ onOpenChange }) {
             }}
             disabled={submitting}
             autoComplete="current-password"
+            aria-describedby={error ? "delete-account-error" : undefined}
+            aria-invalid={error ? true : undefined}
           />
         </div>
 
         {error && (
-          <p className="text-sm text-destructive" role="alert">
+          <p
+            id="delete-account-error"
+            className="text-sm text-destructive"
+            role="alert"
+          >
             {error}
           </p>
         )}
@@ -121,7 +109,6 @@ function DeleteAccountForm({ onOpenChange }) {
           type="submit"
           variant="destructive"
           disabled={submitting || !password}
-          onClick={handleConfirm}
         >
           {submitting && (
             <Loader2 className="h-4 w-4 animate-spin" aria-hidden="true" />
--- a/frontend/src/components/Profile/DeleteAccountDialog.jsx
+++ b/frontend/src/components/Profile/DeleteAccountDialog.jsx
@@ -4,7 +4,6 @@ import { Loader2 } from "lucide-react";
 
 import {
   AlertDialog,
-  AlertDialogAction,
   AlertDialogCancel,
   AlertDialogContent,
   AlertDialogDescription,
@@ -12,6 +11,7 @@ import {
   AlertDialogHeader,
   AlertDialogTitle,
 } from "@/components/ui/alert-dialog";
+import { Button } from "@/components/ui/button";
 import { Input } from "@/components/ui/input";
 import { Label } from "@/components/ui/label";
 import { deleteAccount } from "@/services/userService";
@@ -105,7 +105,15 @@ function DeleteAccountForm({ onOpenChange }) {
         <AlertDialogCancel disabled={submitting} type="button">
           Cancel
         </AlertDialogCancel>
-        <AlertDialogAction
+        {/*
+         * NOTE: This is a plain <Button>, not an <AlertDialogAction>.
+         * Radix's AlertDialogAction wraps Dialog.Close, which calls
+         * onOpenChange(false) on click — that would unmount the form
+         * before our async API call resolves, hiding the loading state
+         * and any inline error. We close the dialog ourselves only after
+         * a successful deletion (in handleConfirm).
+         */}
+        <Button
           type="submit"
           variant="destructive"
           disabled={submitting || !password}
@@ -114,7 +122,7 @@ function DeleteAccountForm({ onOpenChange }) {
             <Loader2 className="h-4 w-4 animate-spin" aria-hidden="true" />
           )}
           {submitting ? "Deleting…" : "Delete my account"}
-        </AlertDialogAction>
+        </Button>
       </AlertDialogFooter>
     </form>
   );
--- a/frontend/src/components/Profile/DeleteAccountDialog.jsx
+++ b/frontend/src/components/Profile/DeleteAccountDialog.jsx
@@ -1,0 +1,146 @@
+import { useState } from "react";
+import { useNavigate } from "react-router-dom";
+import { Loader2 } from "lucide-react";
+
+import {
+  AlertDialog,
+  AlertDialogAction,
+  AlertDialogCancel,
+  AlertDialogContent,
+  AlertDialogDescription,
+  AlertDialogFooter,
+  AlertDialogHeader,
+  AlertDialogTitle,
+} from "@/components/ui/alert-dialog";
+import { Input } from "@/components/ui/input";
+import { Label } from "@/components/ui/label";
+import { deleteAccount } from "@/services/userService";
+import { clear as clearTokens } from "@/services/tokenStore";
+import { useToast } from "@/hooks/useToast";
+
+// Body lives in a separate component so its state resets on every open/close
+// cycle: Radix unmounts AlertDialogContent on close, remounting this fresh.
+function DeleteAccountForm({ onOpenChange }) {
+  const navigate = useNavigate();
+  const { toast } = useToast();
+  const [password, setPassword] = useState("");
+  const [deleteStories, setDeleteStories] = useState(false);
+  const [submitting, setSubmitting] = useState(false);
+  const [error, setError] = useState("");
+
+  async function handleConfirm(e) {
+    e.preventDefault();
+    if (!password) {
+      setError("Please enter your password to confirm.");
+      return;
+    }
+    setSubmitting(true);
+    setError("");
+    try {
+      await deleteAccount(password, deleteStories);
+      // tokenStore.clear() dispatches an "auth:logout" event that the
+      // AuthContext listens for to drop the in-memory user.
+      clearTokens();
+      onOpenChange(false);
+      toast.success("Your account has been deleted.");
+      navigate("/");
+    } catch (err) {
+      const data = err?.response?.data;
+      const message =
+        (Array.isArray(data?.password) && data.password[0]) ||
+        data?.detail ||
+        err?.message ||
+        "Failed to delete account. Please try again.";
+      setError(message);
+      setSubmitting(false);
+    }
+  }
+
+  return (
+    <form onSubmit={handleConfirm} noValidate>
+      <AlertDialogHeader>
+        <AlertDialogTitle>Delete your account?</AlertDialogTitle>
+        <AlertDialogDescription>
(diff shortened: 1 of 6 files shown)


sezindogan (changes requested):
## Overview                  
                                                                                                                                                       
  Adds a "Delete account" flow to the frontend: a DeleteAccountDialog (password confirmation + optional story hard-delete) wired into a new "Danger zone" section at the bottom of EditProfileForm. The service call hits DELETE /users/me/ with password, hard_delete, and refresh in the body. Test coverage is solid.                                                                                                                                   
  
## Issues
                                                                                                                                                                  
 ### Critical issue — double handleConfirm invocation
                                                                                                                                                       
  In DeleteAccountDialog.jsx, the confirm button is both inside a` <form onSubmit={handleConfirm}> `and has its own `onClick={handleConfirm}`:
                                                                                                                                                       

 // DeleteAccountDialog.jsx  lines ~114–124
```
  <AlertDialogAction                                                                                                                                   
    type="submit"           
    ...                                                                                                                                                
    onClick={handleConfirm}   // fires on click
  >                                                                                                                                                    
  
```
  type="submit" causes the button click to bubble up and trigger the form's onSubmit, so handleConfirm is called twice on every confirmation. The first call sets submitting = true, but since the state update is async, the second call can race through the if (!password) guard and fire a second api.delete. Fix: remove onClick={handleConfirm} from the button and let the form's onSubmit be the single handler.                                   
                            
###  Major behavioural discrepancy vs mobile
                                                                                                                                                       
  Frontend (this PR): defaults to hard_delete: false — stories are anonymised; opt-in checkbox to delete them.
                                                                                                                                                       
  Mobile (ProfileScreen.tsx:2019): always passes hard_delete: true — stories are always permanently deleted with no choice given. The modal text confirms this: "Your stories, comments, and likes will also be permanently deleted." This is what the backend supports as well                                                                
                                                                                                                                                       
  These two clients present completely opposite defaults to the user. Someone who deletes via the web app expects their stories to survive (anonymised); on mobile they get silently wiped. Align with mobile implementation.
                                                                                                                                                                             
##  Other issues
              
 1. deleteStories vs hard_delete naming leak
                                                                                                                                                       
  The service function signature deleteAccount(password, deleteStories) maps deleteStories to hard_delete, but the parameter name carries frontend semantics into the service layer. The mobile data source names the same parameter hardDelete (consistent with the API field). Consider renaming to hardDelete in the frontend service too for consistency.                                                                                              
                          
 2. Auth state not fully cleared

  clearTokens() is documented to dispatch auth:logout so AuthContext drops the in-memory user. But if anything in the render tree holds a reference to the user object (e.g. cached route state), the user could briefly see stale UI before the navigate to /. Low risk since navigate("/") follows immediately, but worth a smoke-test.                                                                                                                 
                            
 3. Missing data-testid or aria-label on password input                                                                                                  
   
  The test resolves the input via getByLabelText(/confirm with your password/i), which works, but the <Input> lacks aria-describedby pointing to the error paragraph. Screen-reader users won't hear the inline error unless the role="alert" is announced separately. Consider adding aria-describedby="delete-account-error" and a matching id on the error element.          
 
4. Delete account button location inconsistent with mobile                                                                                              
                                                                                                                                                       
  On mobile, "Delete Account" is surfaced directly on the profile view (alongside "Sign out" and "Reset password") without entering edit mode. On the frontend it is buried inside the edit form under the "Danger zone" section, requiring the user to click "Edit profile" and scroll to the bottom. The two clients should align on a single UX pattern — mobile's placement (account-level actions grouped outside of editing) is the more standard convention.                                                           
                            
##  Code quality
              
  - State reset logic (the inner DeleteAccountForm component unmounted by Radix on close) is clever and well-commented.
  - Test suite is comprehensive: happy path, both hard_delete values, password validation, loading state, error variants. Good coverage.               
  - The Boolean(deleteStories) cast in userService.js is a safe guard but callers already pass a boolean — minor.

omarreis01:
Thanks for the thorough review @sezindogan! Pushed [`8c1f326`](https://github.com/bounswe/bounswe2026group4/commit/8c1f326) addressing the actionable items:

### Fixed

- **Critical — double `handleConfirm` invocation.** Removed `onClick={handleConfirm}` from `AlertDialogAction`; the form's `onSubmit` is now the single handler, so the click no longer bubbles into a second `handleConfirm` call. Added a `toHaveBeenCalledTimes(1)` assertion to lock this in.
- **Major — alignment with mobile (always hard-delete).** Dropped the opt-in checkbox and the "stories will be anonymised" copy. The dialog now always passes `hard_delete: true` and tells the user "Your stories, comments, and likes will also be permanently deleted." — matching `ProfileScreen.tsx` exactly.
- **Naming leak — `deleteStories` → `hardDelete`.** Renamed the service parameter so the frontend matches the backend field name and the mobile data source (`profileRemoteSource.deleteAccount`). Also dropped the now-unnecessary `Boolean(...)` cast.
- **A11y — `aria-describedby` on password input.** The password `Input` now sets `aria-describedby="delete-account-error"` and `aria-invalid={true}` whenever an inline error is present, and the error `<p>` carries the matching `id`. Screen readers will now announce the validation message together with the field. Added a test that asserts both attributes after a 400 response.

### Not changed (intentionally)

- **Auth state freshness.** `clearTokens()` dispatches `auth:logout`, and `navigate("/")` runs in the same tick, so any stale-UI window is effectively zero in practice — agreed it's low risk and didn't seem worth reworking the flow for it.
- **Delete button placement.** Kept the trigger inside the "Danger zone" section of `EditProfileForm` for now. The web ProfilePage doesn't currently surface "Sign out" / "Reset password" the way the mobile profile screen does, so moving only "Delete account" up to the top would be inconsistent in a different way. Happy to coordinate a separate follow-up to align both clients (and surface those other actions on web) if the team wants to standardise.

All 927 frontend tests pass; lint and build are clean.

sezindogan:
The four fixes from the previous round look correct on inspection — double-invocation guard, hard-delete alignment, parameter rename, and aria-describedby. However, there is another critical bug I just detected during manual testing.                                                                
                                                                                                                                                                         
###  Critical — confirm button does nothing (regression from the double-invocation fix)
                                                                                                                                                       
  Root cause: AlertDialogAction is not a plain button — Radix UI implements it as DialogPrimitive.Close. The moment the user clicks it, Radix fires onOpenChange(false) synchronously, which collapses the {open && <DeleteAccountForm …/>} guard in the parent and unmounts the form before onSubmit can fire. handleConfirm never runs. The user sees the dialog close with no API call made.
                                                                                                                                                       
  The previous version had onClick={handleConfirm} on AlertDialogAction, which is why the API call did fire (the click handler ran before Radix closed the dialog). Removing it silenced the double-invocation but also silenced the only working code path.
                                                                                                                                                       
  Fix: Replace AlertDialogAction with a regular Button (destructive variant, type="submit") that has no built-in close behaviour. handleConfirm already calls onOpenChange(false) on success and leaves the dialog open on error, so the close lifecycle is already managed correctly — the submit button just needs to stop triggering it prematurely.                                                                                                        
                            
  // replace AlertDialogAction with:
```
  <Button type="submit" variant="destructive" disabled={submitting || !password}>
    {submitting && <Loader2 className="h-4 w-4 animate-spin" aria-hidden="true" />}                                                                    
    {submitting ? "Deleting…" : "Delete my account"}                                                                                                   
  </Button>      
```                                                                                                                                      
                                                                                                                                                       
  Also remove AlertDialogAction from the import list and add Button.                                                                                   
   
  Test to add: After clicking confirm, assert that deleteAccount was called — the existing toHaveBeenCalledTimes(1) test already does this, but it currently passes vacuously because the mock resolves before the unmount race is observable in jsdom. Add a test that asserts the button click does not close the dialog when the request is pending (i.e. onOpenChange not called until the promise resolves).

omarreis01:
Quick follow-up — went back through the dialog after a deeper read of Radix's source and found that the prior fix in [`8c1f326`](https://github.com/bounswe/bounswe2026group4/commit/8c1f326) introduced a subtle regression. Pushed [`45770af`](https://github.com/bounswe/bounswe2026group4/commit/45770af) to fix it.

### What I missed the first time

`AlertDialogAction` is just a wrapper over `Dialog.Close` from `@radix-ui/react-dialog`:

```js
// node_modules/@radix-ui/react-dialog/dist/index.mjs
var DialogClose = React.forwardRef((props, forwardedRef) => {
  // ...
  return jsx(Primitive.button, {
    type: "button",
    ...closeProps,
    ref: forwardedRef,
    onClick: composeEventHandlers(props.onClick, () => context.onOpenChange(false))
  });
});
```

So `AlertDialogAction` always calls `onOpenChange(false)` on click — unless the consumer's own `onClick` calls `e.preventDefault()`, because [`composeEventHandlers`](https://github.com/radix-ui/primitives/blob/main/packages/react/primitive/src/primitive.tsx) skips the second handler when `event.defaultPrevented` is true.

That's actually what the original `onClick={handleConfirm}` was doing (likely unintentionally): `handleConfirm` calls `e.preventDefault()` first thing, which both (a) cancelled the implicit form-submit triggered by the click and (b) short-circuited Radix's auto-close. After I removed `onClick`, neither of those was happening anymore — clicking the button would fire `onOpenChange(false)` on the click, the parent (`EditProfileForm`) would set its `deleteDialogOpen` state to false, and the form would unmount before `await deleteAccount(…)` ever resolved. In production this means:

- the **"Deleting…" loading state never appears** (form unmounts immediately), and
- **backend errors like "Incorrect password" are silently swallowed** — the user sees the dialog vanish and… nothing happens. Way worse than the original double-invocation theory.

The existing tests didn't catch it because `renderDialog` hard-codes `open={true}` and uses a `vi.fn()` for `onOpenChange`, so the dialog never actually closes in tests.

### The fix

Replaced `AlertDialogAction` with a plain `<Button type="submit" variant="destructive">`. Now:

- Click on the destructive button → browser fires `submit` on the form → `handleConfirm` runs (single invocation, no Radix interference).
- Enter in the password input → same path.
- The dialog only closes when `handleConfirm` explicitly calls `onOpenChange(false)` after a successful deletion. Errors stay inline and the form stays mounted.

I also wrote three regression guards under a new `with controlled \`open\` (production wiring)` describe block that uses a real `useState` harness (matching `EditProfileForm`'s setup):

1. `keeps the dialog mounted while the request is in flight` — confirms the loading state is visible.
2. `keeps the dialog open and surfaces inline error on backend failure` — confirms the 400 path still shows the error.
3. `submits via Enter key in the password input` — confirms keyboard submission still routes through `handleConfirm`.

I verified that test 1 and test 2 actually fail against [`8c1f326`](https://github.com/bounswe/bounswe2026group4/commit/8c1f326) before the fix and pass after, so they're real guards, not just coverage padding.

All 930 frontend tests pass; lint and build clean.

sezindogan:
The issue is now fixed. There is one remaining inconsistency.

EditProfileForm.jsx:305 still reads:                                                                                                                 
'Permanently delete your account and (optionally) your stories.'                                                                                       
The "optionally" is stale — the checkbox was removed and deletion is now always hard-delete. Should be updated to something like:                    
'Permanently delete your account, stories, comments, and uploads.'      

After this, it is ready for merge.

omarreis01:
Good catch, thanks @sezindogan — pushed [`235c1bd`](https://github.com/bounswe/bounswe2026group4/commit/235c1bd) updating [EditProfileForm.jsx:305](frontend/src/components/Profile/EditProfileForm.jsx#L305) to drop the stale "(optionally)" wording. The Danger zone helper now reads:

> Permanently delete your account, stories, comments, and uploads.

which matches both the dialog copy and the actual behaviour. Lint, tests, and build all clean.

sezindogan (approved):
Ready for merge.

PR #591, opened 2026-05-08
feat(frontend): timeline page mobile parity filters + radiogroup keyboard nav

### LINKED ISSUE #543: [Feature][Frontend] Timeline page: mobile parity filters + radiogroup keyboard nav
## Description
Follow-up to PR #538 (frontend timeline page). The reviewer explicitly requested a tracking issue so these items don't get lost.

### 1. Mobile parity gaps — search / location / tag / active-chip filters
The timeline page currently only exposes the time-window filter. Mobile's `TimelineScreen` also offers text search, location filtering, tag filtering, and visible active-filter chips. The frontend already has reusable components for all of these (used on Feed/Map pages); they should be wired into the timeline page.

- [ ] **Text search bar** — reuse `SearchFilter` / `SearchBar` (as on Feed page). Mobile renders `StorySearchControls helperText=\"Search by title or place.\"` above the period selector.
- [ ] **Location filter (Nominatim geocoding)** — `timelineService.js` already passes `lat_min/lat_max/lng_min/lng_max` through; no UI populates them yet. Plug into the Nominatim geocoder being added in PR #531 (don't duplicate the geocoder).
- [ ] **Tag filter** — wire existing `TagFilterInput.jsx` + tag-search service into the timeline filter state, and add a `tags` param to `getTimeline`.
- [ ] **Active filter chips** — reuse existing `ActiveFilters.jsx` to show dismissible chips for active location / time range / tags below the search bar.

> Coordination note: blocked on / coordinated with PR #531 (Nominatim geocoding in `FilterPanel`). Once that lands, the bbox output maps directly to the existing passthrough params in `timelineService.js`.

### 2. Timeline mode selector — roving tabindex for radiogroup (a11y polish)
PR #538 switched the four mode pills to `role=\"radiogroup\"` + `role=\"radio\"` + `aria-checked`. The WAI-ARIA authoring spec for `radiogroup` expects a **roving tabindex** — only the checked radio is in the tab order, and arrow keys move between options. Currently each pill is independently tab-focusable. Functionally fine for mouse/touch users; deviates from the pattern for keyboard users.

- [ ] Implement roving tabindex on the four mode pills (only the checked one is tabbable; arrow keys cycle).

## Acceptance Criteria
- [ ] Timeline page exposes a text search bar matching mobile UX
- [ ] Timeline page supports location filtering via the shared Nominatim geocoder (after #531 lands)
- [ ] Timeline page supports tag multi-select; `tags` param flows through `getTimeline`
- [ ] Active filters render as dismissible chips on the timeline page
- [ ] Mode selector implements roving tabindex per WAI-ARIA radiogroup pattern
- [ ] Tests cover the new filter wiring and keyboard navigation

## Context
- Source review: https://github.com/bounswe/bounswe2026group4/pull/538
- Mobile reference: PR #527 (`TimelineScreen`)
- Geocoder dependency: PR #531

omarreis01:
# Description
Fixes #543

Brings the desktop **Timeline page** in line with mobile's `TimelineScreen` feature set and lands the WAI-ARIA *roving-tabindex* behaviour PR [#538](https://github.com/bounswe/bounswe2026group4/pull/538) deferred. All four bullets in the issue are addressed:

- **Text search bar** — `SearchBar` is rendered above the time-window card with the mobile placeholder *"Search by title or place…"*. The `q` URL param is the source of truth.
- **Location filter** — reuses the Nominatim-backed `FilterPanel` (added in PR [#531](https://github.com/bounswe/bounswe2026group4/pull/531)). To avoid duplicating the time-window UI, `FilterPanel` gains a small new `hideYearRange` prop (default `false`) — only Timeline opts in.
- **Tag multi-select** — `TagFilterInput` (already inside `FilterPanel`) now flows through to `getTimeline`.
- **Active filter chips** — `ActiveFilters` renders dismissible chips for `q` / location / proximity / tags. Year is *not* surfaced as a chip because the time-window selector already represents it visually, and rendering both would be redundant.
- **Roving tabindex on mode pills** — only the currently-checked radio is in the tab order; `ArrowLeft/Right/Up/Down` move focus *and* selection (with wrap-around); `Home`/`End` jump to the ends.

## Backend audit (verified before / during implementation)

Every assumption the frontend makes was checked against `backend/apps/stories/{views,serializers,services}.py` and `backend/common/pagination.py`:

| Concern | Backend reality | How the PR handles it |
|---|---|---|
| `/stories/timeline/` accepted params | `year_from`, `year_to`, `lat_min/max`, `lng_min/max`, `has_image`, `page`, `page_size` only — validated by `TimelineQuerySerializer`. No `q`, `tags`, free-text `location`, or proximity. | `getTimeline` proxies `/stories/timeline/` directly when the active filter set fits this shape. |
| `/stories/feed/` & `/stories/search/` accepted params | Inherit `FeedQuerySerializer`: `q` (required for search, `min_length=1`, whitespace-stripped), `sort_by`, `year_from/to`, `location`, `tags[]`, `latitude/longitude/radius_km`, `lat_min/max/lng_min/max`. Validated to require lat/lng/radius_km **together**; bbox four-together; `radius_km ∈ [0.001, 500]`. | Fallback only sends the lat/lng/radius_km triple together. Frontend `radiusKm` options (0.5 / 1 / 10 / 100) all sit inside `[0.001, 500]`. `q` is trimmed before being sent. |
| `tags` wire format | DRF `ListField(child=CharField)` reads repeated keys: `?tags=foo&tags=bar`. Backend filter is `name__iexact`, AND-chained — each named tag must be on the story. | `api.js` paramsSerializer emits arrays as repeated keys, so `params.tags = ["foo","bar"]` lands as `?tags=foo&tags=bar`. Tag *names* flow straight through from `useFilterState`'s URL string — no ID lookup needed. |
| `q` semantics | `Q(title__icontains) | Q(location_name__icontains)` — title OR place. | Placeholder *"Search by title or place…"* matches the actual behaviour (and the mobile copy). |
| Pagination cap | `StoryPagination.max_page_size = 100`. | `FALLBACK_FETCH_PAGE_SIZE = 100` — the ceiling. Stories beyond the 100th match in fallback won't appear; same cap mobile uses. |
| **Serializer shape delta in fallback** | `StoryTimelineSerializer` returns `photo_url` + `temporal_coverage` (EDTF) but no `location_name`. `StoryFeedSerializer` returns the inverse: `location_name`, plus feed-card extras, but **no** `photo_url` or `temporal_coverage`. | **Cosmetic UX delta in fallback mode.** `TimelineEntry` already conditionally renders both `photo_url` and `temporal_coverage`, so no crash — fallback cards just show no thumbnail and no EDTF chip, but gain a place name. **Mobile has the exact same limitation** (its mapper looks for `media_items[]`, which feed/search don't expose either). Platform-consistent, not a bug. |

### Fallback strategy in `timelineService.js`

- **All-supported filters?** → call `/stories/timeline/` directly (server-side ordering preserved).
- **Any unsupported filter?** → call `/stories/search/` (with `q`) or `/stories/feed/`, request up to `FALLBACK_FETCH_PAGE_SIZE = 100` results, re-sort by `getTimelineHistoricalYear(story)`, paginate client-side.

`getTimelineHistoricalYear` is exported and unit-tested per `time_type` so the formula stays in sync with the server's CASE-WHEN in `apps/stories/services.py::get_story_timeline`.

# Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Documentation

# Screenshots / Recordings

# Estimated Review Time
- [ ] < 5 min
- [ ] 5-15 min
- [x] > 15 min

---

### Notes for reviewer

- **Why not modify `/stories/timeline/` to accept `q`/`tags`?** That's a backend change and out of scope for this frontend issue; mobile uses the same fallback so the cross-platform behaviour matches.
- **Why a new `hideYearRange` prop on `FilterPanel` instead of a new `TimelineFilterPanel` component?** The two would have been near-clones — `hideYearRange` is one extra prop and one conditional render. Default is `false`, so Feed and Map are untouched.
- **Year isn't a chip.** The time-window selector card visually owns the year filter. Rendering "1875" both as a chip and as a Year-mode card was deemed noise; happy to add chips if reviewers prefer.
- **Photo / EDTF chip drop-out in fallback mode** — see audit table above. Same as mobile; documented as a known cosmetic delta, not a regression.
- **All 992 frontend tests pass** locally; lint and `npm run build` are green.

### DIFF
--- a/frontend/src/components/SearchFilter/ActiveFilters.jsx
+++ b/frontend/src/components/SearchFilter/ActiveFilters.jsx
@@ -44,7 +44,7 @@ function TagChip({ name, onRemove }) {
  * Renders removable chips for each active search/filter value.
  * Returns null when no filters are active.
  *
- * onRemove receives a key: "q" | "year_range" | "year_from" | "year_to" | "location"
+ * onRemove receives a key: "q" | "year_range" | "year_from" | "year_to" | "location" | "proximity" | "has_image"
  * onRemoveTag receives a tag name string
  */
 function ActiveFilters({
@@ -55,6 +55,7 @@ function ActiveFilters({
   radiusKm = null,
   hasProximity = false,
   tags = [],
+  hasImage = false,
   onRemove,
   onRemoveTag,
   onClearAll,
@@ -80,6 +81,10 @@ function ActiveFilters({
   if (hasProximity && radiusKm != null) {
     const distanceLabel = radiusKm < 1 ? `${Math.round(radiusKm * 1000)} m` : `${radiusKm} km`;
     chips.push({ key: "proximity", label: `Within ${distanceLabel}` });
+  }
+
+  if (hasImage) {
+    chips.push({ key: "has_image", label: "Has image" });
   }
 
   if (chips.length === 0 && tags.length === 0) return null;
--- a/frontend/src/components/SearchFilter/FilterPanel.jsx
+++ b/frontend/src/components/SearchFilter/FilterPanel.jsx
@@ -28,7 +28,7 @@ const PROXIMITY_OPTIONS = [
  * The parent should pass a `key` tied to the current filter values so that
  * the component resets its local form whenever filters are cleared externally.
  */
-function FilterPanel({ yearFrom = "", yearTo = "", location = "", latMin = null, latMax = null, lngMin = null, lngMax = null, latitude = null, longitude = null, radiusKm = null, tags = [], onApply, activeCount = 0 }) {
+function FilterPanel({ yearFrom = "", yearTo = "", location = "", latMin = null, latMax = null, lngMin = null, lngMax = null, latitude = null, longitude = null, radiusKm = null, tags = [], onApply, activeCount = 0, hideYearRange = false }) {
   const [open, setOpen] = useState(false);
   const [localYearFrom, setLocalYearFrom] = useState(yearFrom);
   const [localYearTo, setLocalYearTo] = useState(yearTo);
@@ -267,6 +267,7 @@ function FilterPanel({ yearFrom = "", yearTo = "", location = "", latMin = null,
         >
           <div className="space-y-4">
             {/* Year range */}
+            {!hideYearRange && (
             <fieldset>
               <legend className="mb-2 text-sm font-medium">Year range</legend>
               <div className="flex items-center gap-2">
@@ -344,6 +345,7 @@ function FilterPanel({ yearFrom = "", yearTo = "", location = "", latMin = null,
                 </p>
               )}
             </fieldset>
+            )}
 
             {/* Location */}
             <div>
--- a/frontend/src/components/SearchFilter/FilterPanel.jsx
+++ b/frontend/src/components/SearchFilter/FilterPanel.jsx
@@ -28,13 +28,14 @@ const PROXIMITY_OPTIONS = [
  * The parent should pass a `key` tied to the current filter values so that
  * the component resets its local form whenever filters are cleared externally.
  */
-function FilterPanel({ yearFrom = "", yearTo = "", location = "", latMin = null, latMax = null, lngMin = null, lngMax = null, latitude = null, longitude = null, radiusKm = null, tags = [], onApply, activeCount = 0, hideYearRange = false }) {
+function FilterPanel({ yearFrom = "", yearTo = "", location = "", latMin = null, latMax = null, lngMin = null, lngMax = null, latitude = null, longitude = null, radiusKm = null, tags = [], hasImage = false, onApply, activeCount = 0, hideYearRange = false, showHasImage = false }) {
   const [open, setOpen] = useState(false);
   const [localYearFrom, setLocalYearFrom] = useState(yearFrom);
   const [localYearTo, setLocalYearTo] = useState(yearTo);
   const [localLocation, setLocalLocation] = useState(location);
   const [suggestionsQuery, setSuggestionsQuery] = useState("");
   const [localTags, setLocalTags] = useState(tags);
+  const [localHasImage, setLocalHasImage] = useState(hasImage);
   const [yearError, setYearError] = useState("");
   const [lockedBbox, setLockedBbox] = useState(
     () => latMin != null && latMax != null && lngMin != null && lngMax != null
@@ -201,7 +202,8 @@ function FilterPanel({ yearFrom = "", yearTo = "", location = "", latMin = null,
       latitude: proximityActive ? localCoords.latitude : null,
       longitude: proximityActive ? localCoords.longitude : null,
       radiusKm: proximityActive ? localRadiusKm : null,
-      tags: localTags
+      tags: localTags,
+      ...(showHasImage ? { hasImage: localHasImage } : {}),
     });
     setOpen(false);
   }
@@ -218,6 +220,7 @@ function FilterPanel({ yearFrom = "", yearTo = "", location = "", latMin = null,
     setProximityResolving(false);
     setProximityStatus(null);
     setLocalTags([]);
+    setLocalHasImage(false);
     setYearError("");
     clearSuggestions();
     onApply({
@@ -232,6 +235,7 @@ function FilterPanel({ yearFrom = "", yearTo = "", location = "", latMin = null,
       longitude: null,
       radiusKm: null,
       tags: [],
+      ...(showHasImage ? { hasImage: false } : {}),
     });
     setOpen(false);
   }
@@ -464,6 +468,22 @@ function FilterPanel({ yearFrom = "", yearTo = "", location = "", latMin = null,
 
             {/* Tags */}
             <TagFilterInput value={localTags} onChange={setLocalTags} />
+
+            {showHasImage && (
+              <fieldset>
+                <legend className="sr-only">Image presence</legend>
+                <label className="flex cursor-pointer items-center gap-2 text-sm">
+                  <input
+                    type="checkbox"
+                    checked={localHasImage}
+                    onChange={(e) => setLocalHasImage(e.target.checked)}
+                    aria-label="Only stories with an image"
+                    className="h-4 w-4 rounded border-input"
+                  />
+                  <span>Only stories with an image</span>
+                </label>
+              </fieldset>
+            )}
 
             {/* Actions */}
             <div className="flex items-center justify-between gap-2 pt-1">
--- a/frontend/src/components/SearchFilter/__tests__/ActiveFilters.test.jsx
+++ b/frontend/src/components/SearchFilter/__tests__/ActiveFilters.test.jsx
@@ -181,4 +181,23 @@ describe("ActiveFilters", () => {
 
     expect(onRemove).toHaveBeenCalledWith("proximity");
   });
+
+  it("renders a 'Has image' chip when hasImage is true", () => {
+    renderFilters({ hasImage: true });
+    expect(screen.getByText(/has image/i)).toBeInTheDocument();
+  });
+
+  it("does not render the 'Has image' chip when hasImage is false / unset", () => {
+    renderFilters({});
+    expect(screen.queryByText(/has image/i)).not.toBeInTheDocument();
+  });
+
+  it("calls onRemove with 'has_image' when the Has-image chip X is clicked", async () => {
+    const user = userEvent.setup();
+    const onRemove = vi.fn();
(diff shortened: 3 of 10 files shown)


sezindogan:
Overview                                                                                                                                             
                                                                                                                                                       
  Adds search (q), location/bbox, proximity, and tags filtering to the Timeline page, a WAI-ARIA roving-tabindex radiogroup for the time window selector, and a fallback path in timelineService.js for filter combinations /stories/timeline/ doesn't support (routes to /stories/search/ or /stories/feed/, then client-sorts and paginates).                                                                                                    
                            

 ### Design Change Required: Remove the Banner
                                           
  The "Choose a time window" card should be removed. Timeline should use the same SearchBar + FilterPanel + ActiveFilters pattern as Feed and Map, with year range inside the filter tab.                                                                                                                   
   
  What to remove:                                                                                                                                      
  - The <Card> "Choose a time window" block and all state that serves it: mode, yearInput, rangeFrom, rangeTo, deriveYearFilter, MODES, MODE_LABELS,
  handleModeKeyDown, modeRefs, countActiveFilters, the status row.                                                                                     
  - hideYearRange prop on FilterPanel (and its implementation) — becomes unnecessary.
  - The hideYearRange tests in FilterPanel.test.jsx become obsolete.                                                                                   
  - Year inputs should be wired through useFilterState / URL params like the other filters so they survive refresh.                                    
                                                                                                                   
  Note: mobile currently also uses an inline TimelinePeriodSelector, so this change intentionally diverges from mobile and a follow-up should align it.
                                                                                                                                                                                                                                                                                                     
 ### Missing Feature: Filter by Image Presence                                                                                                            
                                           
  Since this PR is about timeline filtering, a "Has image" toggle belongs here rather than in a separate issue/PR.
                                                                                                                                                       
  What's needed:
  - Add has_image param to getTimeline, forwarded on the primary path and applied client-side on the fallback (results.filter(s => has_image ?         
  s.has_image : true)).                                                                                                                                
  - Add a "Has image" toggle to FilterPanel, gated behind a prop (Feed/Map don't need it).
  - Surface it as an ActiveFilters chip and wire it through useFilterState.                                                                            
  - Cover it in timelineService.test.js and TimelinePage.test.jsx.                                                                                     
                                                                                                                                                  
  ## Issues                    
                                                                                                                                                       
 1. Fallback cap silently truncates results (timelineService.js:900)
  FALLBACK_FETCH_PAGE_SIZE = 100 is a hard cap — stories beyond 100 are inaccessible and the returned count misrepresents the true API count. Worth a
  test explicitly documenting this ceiling, and a code comment acknowledging the limitation.                                                           
   
 2. Time window state not URL-serialized                                                                                                                 
  mode, yearInput, rangeFrom, rangeTo live in component state; refreshing loses the time filter while all other filters are preserved in the URL.
  Resolved once the card is replaced by the filter tab.                                                                                                
                                                                                                                                                   
  ## Summary                   
         
  The service layer fallback, filter wiring, and a11y work are solid. Two things block merging: (1) replace the time window card with the shared filter tab, and (2) add image-presence filtering before closing this scope.

omarreis01:
Thanks for the thorough review @sezindogan! Pushed three commits addressing the asks:

### Design change — time-window card removed ✅
[ca520c9](https://github.com/bounswe/bounswe2026group4/pull/591/commits/ca520c9) drops the "Choose a time window" card and rewrites Timeline to use the shared `SearchBar` + `FilterPanel` + `ActiveFilters` pattern, exactly the way Feed and Map do. Removed: the `<Card>` block, `mode` / `yearInput` / `rangeFrom` / `rangeTo` state, `deriveYearFilter`, `MODES` / `MODE_LABELS`, the roving-tabindex `handleModeKeyDown` + `modeRefs`, the status-row label, and `hideYearRange` is no longer used. Year range now lives inside the filter tab and round-trips through the URL like every other filter — refresh-survives, deep-link-shareable. Issue #2 (URL serialization) is resolved by this same change.

> Note: this intentionally diverges from mobile's `TimelinePeriodSelector` per your call-out; happy to file a follow-up to align mobile to this pattern when there's appetite.

### Has-image filter — partial ⚠️
[fe37ea2](https://github.com/bounswe/bounswe2026group4/pull/591/commits/fe37ea2) and [2911605](https://github.com/bounswe/bounswe2026group4/pull/591/commits/2911605):

- `useFilterState` reads/writes `has_image` (URL-serialized).
- `FilterPanel` gains a new `showHasImage` prop (default `false`, so Feed/Map are untouched). When set, the panel renders an "Only stories with an image" checkbox and `onApply` carries `hasImage`.
- `ActiveFilters` renders a removable "Has image" chip with key `has_image`.
- `getTimeline` forwards `has_image=true` to `/stories/timeline/` on the primary path; omitted when false (backend treats absent as "no filter").
- `TimelinePage` opts in via `showHasImage` and wires the chip removal.

**One limitation worth flagging:** `has_image` does **not** apply when the fallback path is active (q / tags / proximity / free-text-location). Reasons:

1. `FeedQuerySerializer` (and `SearchQuerySerializer` which inherits it) does not accept `has_image` as a query param — silently ignored if sent.
2. `StoryFeedSerializer` does not return any image-presence field — no `photo_url`, no `media_items`, no `has_image` flag — so client-side filtering has nothing to read.

Your suggested `results.filter(s => has_image ? s.has_image : true)` would always exclude every story when the toggle is on, because `s.has_image` is always undefined in feed/search responses. The honest fix is a backend addition (annotate `has_image` onto `StoryFeedSerializer` and accept it in `FeedQuerySerializer`), which is outside the frontend team's scope here. I documented the limitation inline in `getTimelineViaFallback`. Tested both branches: `has_image` forwarded on the primary path, omitted on the fallback path.

### Issue #1 — fallback cap visibility ✅
[2911605](https://github.com/bounswe/bounswe2026group4/pull/591/commits/2911605) expands the `FALLBACK_FETCH_PAGE_SIZE` comment to explicitly call out that stories beyond the 100th rank in the underlying endpoint's order are unreachable, and that the returned `count` reflects the fetched window rather than the true backend total. Two new tests:

- "requests at most page_size=100 from the fallback endpoint" — pins the cap.
- "count reflects only stories within the fallback fetch window (cap is observable)" — when the underlying endpoint responds with `count=999, results=[100 stories]`, we surface `count=100` to the page and document why.

### Stats
- 1013 frontend tests pass (+11 vs prior); lint and `npm run build` clean.
- 3 commits, all on `feat/frontend-timeline-mobile-parity` ready for re-review.

sezindogan (approved):
All required changes verified. Code is correct. Tests pass (1013/1013).

###  Review Items — Status

  Design change: Remove time-window card ✅ Fully resolved
  All artifacts removed: the <Card> block, mode / yearInput / rangeFrom / rangeTo state, deriveYearFilter, MODES / MODE_LABELS, the roving-tabindex handleModeKeyDown + modeRefs, and the status row label are all gone from TimelinePage.jsx. Year range is now inside the shared FilterPanel and round-trips through the URL — issue #2 (URL serialization) is resolved by this same change.

  Has-image filter ✅ / ⚠️  Done with an honest backend-bounded limitation
  - Primary path (/stories/timeline/): has_image=true forwarded correctly; omitted when false. Chip renders and is removable. URL-serialized. Wired end-to-end.
  - Fallback path: filter is deliberately a no-op. The reason is sound — StoryFeedSerializer exposes no image flag and FeedQuerySerializer silently drops the param, so client-side filtering would incorrectly exclude everything. The limitation is clearly documented inline in getTimelineViaFallback. This is the right call for a frontend-only PR.

  Issue 1 — fallback cap ✅ Resolved
  FALLBACK_FETCH_PAGE_SIZE now has an explicit multi-line comment acknowledging the truncation and the misrepresented count. Two pinning tests were added: one asserting page_size=100 is the hard cap, and one asserting the returned count reflects the capped window, not the backend total.

  Issue 2 — URL serialization ✅ Resolved (same fix as card removal)

 ### Minor Nit (non-blocking)
  
  hideYearRange is dead code — FilterPanel.jsx:31,274 still accepts and handles the prop, and FilterPanel.test.jsx:496–515 still tests it. No production caller passes it (Feed/Map use SearchFilter, Timeline never passes it). The original review asked for it and its implementation to be removed. It's harmless but leaves the interface slightly wider than it needs to be. 

###  Verdict
  Approve. The two blockers from the first review are resolved. The has_image limitation on the fallback path is an honest constraint that belongs in a
   backend PR, not here, and it's correctly documented. Tests are solid. Ready to merge.

PR #593, opened 2026-05-08
fix(mobile): keep timeline story pins separate from map clusters

### LINKED ISSUE #592: [Bug][Mobile] Map and proximity timeline filters lose story/source context
## Description
Mobile map and search proximity flows can lose the exact source they are meant to represent. The current branch addresses the related issues together because they share the same proximity-filter state and marker rendering path.

Observed problems:
- A story selected from a grouped mobile map marker can be represented only by group coordinates instead of the exact story.
- After zooming or reclustering, the active red timeline pin can be swallowed by a cluster/group marker or highlight the wrong marker.
- The active timeline pin and the currently selected map marker can conflict visually instead of being shown as separate states.
- Group-level timeline actions create ambiguous proximity targets; the timeline action should target an individual story.
- Distance filter chip text can lose useful source context or keep stale map-pin context.
- Applying a proximity filter from the main search controls can leave the user on the timeline tab even though the map is the relevant destination.
- Reapplying a map-pin timeline filter from the filter panel can preserve stale red-pin coordinates instead of resolving the current location and switching to the blue current-location source.

Expected behavior: mobile proximity filters should preserve the exact story or location source, render the active timeline target separately from clusters and current selection, and move the user to the map whenever a map-relevant proximity filter is applied.

## Acceptance Criteria
- [ ] Selecting "View Timeline" from a grouped marker story stores and reuses the target story id.
- [ ] The active map-pin timeline filter renders the chosen story as an individual red pin even when nearby stories are clustered.
- [ ] Zooming in or out does not merge the active timeline story back into a cluster.
- [ ] A currently selected marker can be shown separately from the active timeline pin.
- [ ] Group-level timeline actions are no longer offered for grouped markers; story-specific timeline actions remain available.
- [ ] Distance chips distinguish red map-pin filters from blue current-location filters without stale story/group labels.
- [ ] Applying a proximity filter from the main search controls navigates to the map tab.
- [ ] Reapplying a map-pin timeline filter from the filter panel resolves the current location, converts the source to current location, and keeps the map tab active.
- [ ] Regression tests cover grouped markers, zoom changes, backend grouped markers, story-specific timeline actions, main proximity navigation, and map-pin-to-current-location application.

## Meta (optional)
Effort: M
Priority: P1
Branch: fix/mobile-grouping-and-red-pin

OsmanYusufTosun:
# 📝 Description
Fixes #592

This PR fixes the mobile map behavior where a story selected from a grouped marker timeline flow could be swallowed by clustering and lose its exact red timeline pin. The root cause was that the map-pin timeline filter only carried coordinates/labels, so grouped markers and reclustering could represent the target as a cluster instead of the individual story.

The fix carries the story id through the mobile timeline action and persisted proximity filters, extracts the active timeline story into its own pinned marker before clustering, and keeps that red timeline pin visually distinct from the currently selected marker. It also removes the group-level timeline action so timeline navigation is story-specific, updates the distance chip copy, and expands regression tests around zooming, backend grouped markers, and fallback story matching.

# 📊 Type of Change
- [x]  Bug fix
- [ ]  New feature
- [ ]  Refactoring
- [ ]  Documentation
 
- [ ] Project builds successfully

# 📸 Screenshots / Recordings
N/A

# ⏱️ Estimated Review Time
- [ ] < 5 min
- [ ] 5-15 min
- [x] > 15 min

# ✅ Validation
- `npm test -- --runTestsByPath src/features/map/presentation/screens/__tests__/MapScreen.test.tsx src/app/navigation/__tests__/RootNavigator.test.tsx --runInBand`
  - Passed: 2 test suites, 62 tests
  - Note: existing `VirtualizedList` act warnings appeared in `RootNavigator.test.tsx` output, but the suite passed.

### DIFF
--- a/mobile/src/app/navigation/RootNavigator.tsx
+++ b/mobile/src/app/navigation/RootNavigator.tsx
@@ -731,7 +731,7 @@ export function RootNavigator() {
   };
 
   const handleViewTimelineNearPin = useCallback(
-    (target: { latitude: number; longitude: number; label?: string }) => {
+    (target: { latitude: number; longitude: number; label?: string; storyId?: string }) => {
       updateFilters(
         {
           query: '',
@@ -744,6 +744,7 @@ export function RootNavigator() {
           },
           proximitySource: 'map_pin',
           proximityLabel: target.label,
+          proximityStoryId: target.storyId,
           timeFrom: '',
           timeTo: '',
           tags: [],
--- a/mobile/src/app/navigation/__tests__/RootNavigator.test.tsx
+++ b/mobile/src/app/navigation/__tests__/RootNavigator.test.tsx
@@ -1079,7 +1079,7 @@ describe('RootNavigator auth flow', () => {
         'GET /stories/feed/?page_size=100&sort_by=recent&latitude=41.02&longitude=28.96&radius_km=0.5&page=1',
       );
     });
-    expect(screen.getByLabelText('Remove Distance: 500 m from Harbor Memory red location pin')).toBeTruthy();
+    expect(screen.getByLabelText('Remove Distance: 500 m from red location pin')).toBeTruthy();
 
     fireEvent.press(await screen.findByLabelText('Open timeline story: Harbor Memory'));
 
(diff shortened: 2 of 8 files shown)


MertErenKaplan (approved):
Tested this PR and confirmed the fix works as expected. The mobile map behavior now correctly preserves the selected timeline story, including clustered/grouped marker cases, and the related regression tests pass.

Approving this PR.

PR #596, opened 2026-05-08
feat: display published stories on public profile

### LINKED ISSUE #561: [Feature][Frontend] Display user's stories on their public profile page
### Description

This is the frontend counterpart for issue #551. The backend now exposes a dedicated paginated endpoint for retrieving a user’s published stories:

`GET /users/<user_id>/stories/`

The mobile application should integrate this endpoint into the public profile flow instead of relying solely on `published_story_count`. Public profile screens should display the target user’s published stories using the existing mobile story card components and support pagination/infinite scrolling for scalability.

Since the endpoint already returns a feed-compatible response shape using `StoryFeedSerializer`, the mobile implementation should behave consistently with the existing story feed experience. This includes rendering interaction-aware states such as liked/saved indicators when the requester is authenticated.

The profile story section should only display published stories returned by the endpoint. Draft or removed stories are already filtered server-side and should never appear on mobile.

The implementation should include proper loading, empty, and error states. If the backend returns `404 Not Found` (e.g., inactive or banned user), the mobile app should gracefully handle this case and avoid rendering a broken profile screen.

Pagination should follow the project’s existing feed-loading patterns. The story list should support incremental loading or infinite scroll to avoid rendering excessively large profile payloads at once.

### Acceptance Criteria

- [ ] Public profile screen fetches stories from `GET /users/<user_id>/stories/`
- [ ] Stories are rendered using the existing mobile story card/feed UI
- [ ] Pagination or infinite scrolling is implemented
- [ ] Loading state is displayed while stories are being fetched
- [ ] Empty state is displayed when the user has no published stories
- [ ] Error handling is implemented for failed requests
- [ ] `404 Not Found` responses are handled gracefully for inactive/banned users
- [ ] Authenticated interaction states (`user_has_liked`, `user_has_saved`) are reflected correctly in the UI
- [ ] Stories are ordered consistently with backend response order
- [ ] Existing profile functionality remains unaffected
- [ ] Relevant tests are added or updated

### Meta (optional)

_No response_

sezindogan:
# 📝 Description
Fixes #561 
                                                                                                                                  
  Replace the "Story listing will be available soon." placeholder on the profile page with a fully functional published stories section. The section fetches from GET /users/<user_id>/stories/, renders stories using the existing StoryCard component, and supports pagination. Anonymous stories are excluded server-side and never reach the client.                                                                                                     
                                                            
 ### New features                                                                                                                                         
   
  - PublishedStoriesTab component: paginated list of a user's published stories with loading, empty, error, and 404 states                             
  - getUserStories(userId, { page, pageSize }) added to storyService.js
                                                                                                                                                       
 ### Modified files                                            
                                                                                                                                                       
  - frontend/src/services/storyService.js — added getUserStories                                                                                       
  - frontend/src/pages/ProfilePage.jsx — replaced placeholder with <PublishedStoriesTab />
  - frontend/src/pages/__tests__/ProfilePage.test.jsx — mocked PublishedStoriesTab, replaced outdated "coming soon" test                               
                                                                                                                                                       
###  Tests
                                                                                                                                                       
  - frontend/src/components/Profile/__tests__/PublishedStoriesTab.test.jsx — covers loading state, empty state, story card rendering, total count display, singular/plural label, generic error, 404 error, "Load more" presence/absence, pagination, and correct userId forwarding

# 📊 Type of Change
- [ ]  Bug fix
- [x]  New feature
- [ ]  Refactoring
- [ ]  Documentation
 
# 📸 Screenshots / Recordings

Screenshot 2026-05-08 at 22 15 49

# ⏱️ Estimated Review Time
- [ ] < 5 min
- [x] 5-15 min
- [ ] > 15 min

### DIFF
--- a/frontend/src/components/Profile/PublishedStoriesTab.jsx
+++ b/frontend/src/components/Profile/PublishedStoriesTab.jsx
@@ -1,4 +1,4 @@
-import { useState, useEffect, useCallback } from "react";
+import { useState, useEffect, useCallback, useRef } from "react";
 import { BookOpen, Loader2 } from "lucide-react";
 
 import { Button } from "@/components/ui/button";
@@ -18,27 +18,36 @@ function PublishedStoriesTab({ userId }) {
   const [loadingMore, setLoadingMore] = useState(false);
   const [error, setError] = useState(null);
   const [page, setPage] = useState(1);
-  const [totalCount, setTotalCount] = useState(0);
   const [hasNext, setHasNext] = useState(false);
+  const requestIdRef = useRef(0);
+
+  function handleBookmarkChange(storyId, newBookmarked) {
+    setStories((prev) =>
+      prev.map((s) => (s.id === storyId ? { ...s, user_has_saved: newBookmarked } : s))
+    );
+  }
 
   const fetchStories = useCallback(
     async (pg) => {
+      const id = ++requestIdRef.current;
       if (pg === 1) setLoading(true);
       else setLoadingMore(true);
       setError(null);
       try {
         const data = await getUserStories(userId, { page: pg, pageSize: PAGE_SIZE });
+        if (id !== requestIdRef.current) return;
         setStories((prev) => (pg === 1 ? data.results : [...prev, ...data.results]));
-        setTotalCount(data.count);
         setHasNext(Boolean(data.next));
         setPage(pg);
       } catch (err) {
+        if (id !== requestIdRef.current) return;
         setError(
           is404(err)
             ? "This profile is unavailable or no longer active."
             : "Failed to load published stories. Please try again."
         );
       } finally {
+        if (id !== requestIdRef.current) return;
         setLoading(false);
         setLoadingMore(false);
       }
@@ -77,16 +86,9 @@ function PublishedStoriesTab({ userId }) {
 
   return (
     <div className="space-y-4">
-      <div className="flex items-center gap-2 text-sm text-muted-foreground">
-        <BookOpen className="h-4 w-4" aria-hidden="true" />
-        <span>
-          {totalCount} published {totalCount === 1 ? "story" : "stories"}
-        </span>
-      </div>
-
       <div className="grid grid-cols-1 gap-4 sm:grid-cols-2 lg:grid-cols-3">
         {stories.map((story) => (
-          <StoryCard key={story.id} story={story} />
+          <StoryCard key={story.id} story={story} onBookmarkChange={handleBookmarkChange} />
         ))}
       </div>
 
--- a/frontend/src/components/Profile/PublishedStoriesTab.jsx
+++ b/frontend/src/components/Profile/PublishedStoriesTab.jsx
@@ -47,9 +47,10 @@ function PublishedStoriesTab({ userId }) {
             : "Failed to load published stories. Please try again."
         );
       } finally {
-        if (id !== requestIdRef.current) return;
-        setLoading(false);
-        setLoadingMore(false);
+        if (id === requestIdRef.current) {
+          setLoading(false);
+          setLoadingMore(false);
+        }
       }
     },
     [userId]
--- a/frontend/src/components/Profile/PublishedStoriesTab.jsx
+++ b/frontend/src/components/Profile/PublishedStoriesTab.jsx
@@ -1,0 +1,115 @@
+import { useState, useEffect, useCallback } from "react";
+import { BookOpen, Loader2 } from "lucide-react";
+
+import { Button } from "@/components/ui/button";
+import { ErrorState } from "@/components/ui/error-state";
+import { getUserStories } from "@/services/storyService";
+import StoryCard from "@/components/StoryCard/StoryCard";
+
+const PAGE_SIZE = 10;
+
+function is404(error) {
+  return error?.response?.status === 404;
+}
+
+function PublishedStoriesTab({ userId }) {
+  const [stories, setStories] = useState([]);
+  const [loading, setLoading] = useState(true);
+  const [loadingMore, setLoadingMore] = useState(false);
+  const [error, setError] = useState(null);
+  const [page, setPage] = useState(1);
+  const [totalCount, setTotalCount] = useState(0);
+  const [hasNext, setHasNext] = useState(false);
+
+  const fetchStories = useCallback(
+    async (pg) => {
+      if (pg === 1) setLoading(true);
+      else setLoadingMore(true);
+      setError(null);
+      try {
+        const data = await getUserStories(userId, { page: pg, pageSize: PAGE_SIZE });
+        setStories((prev) => (pg === 1 ? data.results : [...prev, ...data.results]));
+        setTotalCount(data.count);
+        setHasNext(Boolean(data.next));
+        setPage(pg);
+      } catch (err) {
+        setError(
+          is404(err)
+            ? "This profile is unavailable or no longer active."
+            : "Failed to load published stories. Please try again."
+        );
+      } finally {
+        setLoading(false);
+        setLoadingMore(false);
+      }
+    },
+    [userId]
+  );
+
+  useEffect(() => {
+    fetchStories(1);
+  }, [fetchStories]);
+
+  if (loading) {
+    return (
+      <div
+        className="flex items-center justify-center py-12"
+        aria-busy="true"
+        aria-label="Loading published stories"
+      >
+        <Loader2 className="h-6 w-6 animate-spin text-muted-foreground" aria-hidden="true" />
+      </div>
+    );
+  }
+
+  if (error) {
+    return <ErrorState message={error} onRetry={() => fetchStories(1)} />;
+  }
+
+  if (stories.length === 0) {
+    return (
+      <div className="flex flex-col items-center gap-3 py-12 text-muted-foreground">
+        <BookOpen className="h-8 w-8" aria-hidden="true" />
+        <p className="text-sm">No published stories yet.</p>
+      </div>
+    );
+  }
+
+  return (
+    <div className="space-y-4">
+      <div className="flex items-center gap-2 text-sm text-muted-foreground">
+        <BookOpen className="h-4 w-4" aria-hidden="true" />
+        <span>
+          {totalCount} published {totalCount === 1 ? "story" : "stories"}
+        </span>
+      </div>
+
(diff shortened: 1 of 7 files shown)


omarreis01 (commented):
## PR Review: feat: display published stories on public profile

### Summary
Replaces the "Story listing will be available soon." placeholder with a paginated `PublishedStoriesTab` that fetches `GET /users/<id>/stories/`. Implementation is clean, mirrors the existing `SavedStoriesTab` pattern closely, and ships with thorough tests (12 cases covering loading / empty / count / singular-plural / generic error / 404 / pagination / userId forwarding). Overall in good shape — a few small things below, none blocking.

---

### Frontend

#### High
*(none)*

#### Medium

**M1 — Stale-response / no in-flight cancellation in `PublishedStoriesTab.jsx`**
`fetchStories` doesn't track the latest in-flight request. Two concrete scenarios:

1. The user navigates from profile A → B → A quickly. `userId` changes, `useEffect` re-fires `fetchStories(1)`, but if request for B resolves *after* the second request for A, B's results overwrite A's.
2. Inside one profile: while the page-1 fetch is still pending, the user could trigger another retry (e.g. error → click "Try again"). Out-of-order resolution can show the older payload.

A standard fix is an `AbortController` per `userId` (or a `requestId` ref incremented on each call, with `setStories` guarded by `if (id === currentRequestId.current)`). `SavedStoriesTab` has the same shape and the same gap, so this is a pattern issue rather than a regression — but worth fixing now while the component is fresh.

**M2 — Numeric inconsistency between profile header and the new section**
The profile header renders `profile.published_story_count` (counted server-side without filtering on `contributor_visible`, see [`apps/users/services.py:330`](https://github.com/bounswe/bounswe2026group4/blob/main/backend/apps/users/services.py#L330)), but the new section's "X published stories" line uses `data.count` from `/users/<id>/stories/`, which **excludes anonymous stories** for non-owner viewers ([`apps/users/views.py:261-262`](https://github.com/bounswe/bounswe2026group4/blob/main/backend/apps/users/views.py#L261)).

So when User A views User B's profile and B has any `contributor_visible=False` stories, the header may say "7 stories" while the section says "5 published stories" right below it. Looks like a bug to anyone counting.

Easiest frontend-only fix: stop showing the secondary count in the section when it's already displayed in the header (the section's `<BookOpen /> {count} published stories` line is somewhat redundant with the header anyway), or align the wording so the discrepancy is at least explainable.

#### Low

**L1 — Section copy is doubled.**
`<h2>Published Stories</h2>` followed immediately by "X published stories" reads repetitively. Consider just showing the count, or just showing the heading.

**L2 — `StoryCard` is rendered without `onBookmarkChange` here.**
That's the right call (this isn't a bookmarks list), but it means if a logged-in viewer bookmarks/un-bookmarks one of the listed stories from this section, there's no toast and no local state sync — the heart/bookmark icon updates inside `BookmarkButton` itself, but `user_has_saved` on the parent's `stories` state stays stale until a refresh. Not blocking; flagging in case you want a callback that just patches the list item.

**L3 — PR description vs. behavior**
Description says: *"Anonymous stories are excluded server-side and never reach the client."* That's only true for non-owner viewers — owners get their own anonymous stories back, and this section is rendered on the owner's own profile too (it isn't gated on `isOwnProfile` like `SavedStoriesTab` is). That's likely intentional (let the user see all of their own work in one place), but: (a) the PR description undersells what's actually shown, and (b) there's no visual hint in the UI that a card is one the owner posted anonymously, so the owner can't tell which of these are public-facing vs. private-byline. Worth at least updating the description; a small "Anonymous" badge for `contributor_name == null` cards would be a nice follow-up.

**L4 — Potential double-fire on rapid `userId` change.**
Because `fetchStories` is in `useEffect`'s deps and is recreated on every `userId` change, the effect refires correctly — but combined with M1, two near-simultaneous mounts (e.g. React StrictMode double-invoke + a userId change) compound the stale-response problem. A request token / abort is the cleanest fix.

#### Tests
Coverage is genuinely good — loading state, empty state, story card rendering, total count, singular/plural label, generic error, 404 error, "Load more" presence/absence, pagination append, userId forwarding. Two small gaps:

- No test for the **retry path** (error → click "Try again" → success). With `ErrorState.onRetry` wired to `() => fetchStories(1)`, a single test that rejects then resolves on the second call would lock that behavior in.
- No test for the **rapid userId change / stale response** scenario (related to M1). If you address M1, the test for it would be nice.

The "reflects user_has_liked and user_has_saved from story data" test currently only asserts the `data-story-id` attr — the mocked `StoryCard` doesn't render those flags, so the test name is a bit misleading. Consider either renaming it or extending the mock to surface those props.

---

I read the supporting backend code only to verify the API contract — flagging these for the frontend team's awareness, not as PR-blocking items.

- `GET /users/<int:user_id>/stories/` is registered at [`apps/users/profile_urls.py:27`](https://github.com/bounswe/bounswe2026group4/blob/main/backend/apps/users/profile_urls.py#L27) and mounted under `users/` in `config/urls.py`. ✅ matches frontend call.
- `UserStoriesView` uses `StoryPagination` (default `page_size=10`, `page_size_query_param='page_size'`, `max_page_size=100`). The frontend's `{ page, page_size: pageSize }` params line up. ✅
- View permission is `AllowAny`, so the section works for logged-out viewers — good fit for a public profile. ✅
- Service `get_user_published_stories` raises `Http404` for nonexistent or `is_active=False` users, which matches the frontend's `is404(error)` branch. ✅
- `StoryFeedSerializer` returns every field `StoryCard` reads (`id`, `title`, `location_name`, `time_type`, `year`, `year_start`, `year_end`, `preview_text`, `contributor_name`, `user_has_liked`, `user_has_saved`, `tags`). ✅
- Owner-vs-other anonymous filtering is done in the view (M2 above is the consequence of that intentional behavior).
- Backend test coverage in `TestUserStoriesView` is solid — covers unauth, owner vs. non-owner, anonymous filtering, pagination shape, 404 cases, and field presence.

---

- `GET /users/<id>/stories/?page=N&page_size=10` — ✅ endpoint exists, params supported, `AllowAny`.
- Response `{ count, next, previous, results }` — ✅ matches what `getUserStories` returns and what the component destructures.
- `results[*]` shape — ✅ all fields consumed by `StoryCard` are present in `StoryFeedSerializer`.
- 404 on inactive/missing user — ✅ matches frontend's `is404` branch.

---

### Verdict
**APPROVE with minor follow-ups** (no blockers).

Suggested before merge (frontend-only):
- Decide on M2 (the header vs. section count mismatch) — easiest path is dropping the in-section count line.
- Consider M1 (request cancellation / stale-response guard); SavedStoriesTab already has the same pattern, so a tiny shared helper would pay off twice.

Nice-to-have follow-ups (separate PRs are fine):
- L3 — visual indicator for anonymous stories on the owner's own view, plus a PR-description tweak.
- Retry-path test, and a small clarification on the `user_has_liked / user_has_saved` test.

Nice work overall — clean component, real loading/empty/error states, sensible 404 messaging, and the test file is one of the more complete ones I've reviewed in this repo.

sezindogan:
Thanks for the thorough review!

###  Applied fixes:

  - M1 / L4 — Added a requestIdRef guard in fetchStories. Each call increments the ref; state updates are dropped if a newer request has started. This covers both rapid userId changes and the retry-while-pending scenario.
  - M2 / L1 — Removed the {count} published stories line from PublishedStoriesTab. The profile header already shows published_story_count, and the two counts came from different sources (server-side total vs. filtered API response), which was the root of the mismatch. Removed the unused totalCount state as well.
  - Tests — Removed the now-stale count-label tests (shows total count, shows singular 'story'). Added a retry-path test (error → click "Try again" → success). Renamed the user_has_liked / user_has_saved test to passes story data (including user_has_liked and user_has_saved) to StoryCard to accurately describe what the mock actually asserts.

### Also addressed in this PR (L2 and L3):
  
  - L2 — Bookmark state sync: Added handleBookmarkChange in PublishedStoriesTab that patches user_has_saved on the matching story in local state when the bookmark toggle fires. StoryCard already accepted onBookmarkChange — it's now wired up. Added a test: toggle calls the handler and the data-saved attribute on the card updates immediately without a refetch.
  - L3 — Anonymous badge: When contributor_name === null, StoryCard now renders a styled bg-muted chip labelled "Anonymous" instead of unstyled text. Since the backend only returns contributor_name: null for anonymous stories to the owner (non-owners never see those stories), the badge effectively signals to the owner which of their listed stories carry a private byline — no backend changes needed. Updated the StoryCard test to assert the badge element has the bg-muted class.

omarreis01 (approved):
## Re-review — all feedback addressed ✅

Took another pass over the two new commits. Everything from the previous round is resolved cleanly:

- **M1 (stale-response guard)** — `requestIdRef` is incremented per call and guards `setStories` / `setError` / `setLoading` against late resolutions. Correct placement (early-return before state writes, and the `finally` only flips loading flags when the id is still current). Clean fix.
- **M2 (count mismatch)** — the redundant "X published stories" line is gone, so the header's `published_story_count` is now the only count shown. No more inconsistency for viewers of profiles with anonymous stories.
- **L1 (doubled copy)** — resolved with M2.
- **L2 (bookmark sync)** — `handleBookmarkChange` patches `user_has_saved` on the matching item in local state. `StoryCard` is rendered with the callback; toggling stays consistent without a refetch.
- **L3 (anonymous indicator)** — `StoryCard` now renders a styled "Anonymous" pill for `contributor_name == null`. Applied at the `StoryCard` layer, so feed / tags / saved / map popup / timeline all benefit, not just this section. Nice scope choice.
- **Tests** — retry-after-error and bookmark-sync tests added; the anonymous-badge tests in `StoryCard.test.jsx` now verify the element type and `bg-muted` class so a future style regression is caught. The "passes story data to StoryCard" test was also renamed accurately.

### Verification
- CI is green: Backend Tests, Frontend Checks, and Mobile Checks all passed on `58ec456`.
- API contract unchanged from the previous review — still ✅ across the board.

### Findings this round
None blocking. Two micro-notes, both fine to ignore:
- `handleBookmarkChange` is defined inline so its identity changes on every render. `StoryCard` isn't memoized, so this is currently free; mentioning only in case `StoryCard` gets `React.memo`'d later.
- The "anonymous badge" pattern is now duplicated between `StoryCard` and `CommentSection` (which still uses plain text "Anonymous"). Not for this PR, but a tiny shared `<AnonymousBadge />` would be a reasonable follow-up if you want consistency.

LGTM — approving. Thanks for the quick turnaround on the feedback.

PR #599, opened 2026-05-08
feat(frontend/map): View Timeline on pin popup → /nearby-timeline snake view

### LINKED ISSUE #488: [Feature][Frontend] Add 'View Timeline' option to map pins showing nearby stories
## Description
Haritadaki pinlerde mevcut **View Story** seçeneğinin yanına yeni bir **View Timeline** özelliği eklenmeli. Bu butona tıklandığında, seçilen pinin **500 metre çevresindeki** story'ler timeline (zaman çizelgesi) şeklinde gösterilmeli. Timeline tasarımı yılan gibi (snake-like / kıvrımlı) bir görsel akışa sahip olabilir; story'ler kronolojik sırayla bu kıvrımlı çizgi üzerinde sıralanır.

### Behavior
- Map pin popup / detail panelinde **View Story** butonunun yanına **View Timeline** butonu eklenir.
- Tıklandığında o pinin koordinatından 500 m yarıçap içerisindeki tüm story'ler getirilir.
- Story'ler tarihe göre sıralı, snake-like (yılan gibi kıvrılan) bir timeline view'da render edilir.
- Her timeline node'una tıklanınca ilgili story detayına gidilebilir.

## Acceptance Criteria
- [ ] Map pin detail/popup'ına **View Timeline** butonu eklendi
- [ ] Butona tıklayınca pin koordinatının 500 m yarıçapındaki story'ler getiriliyor (backend endpoint mevcut değilse oluşturulmalı)
- [ ] Story'ler kronolojik sırayla snake-like / kıvrımlı bir timeline tasarımında gösteriliyor
- [ ] Timeline üzerindeki her story node'u tıklanabilir ve story detayına yönlendiriyor
- [ ] Yarıçap içinde story yoksa uygun bir empty state gösteriliyor
- [ ] Mobile ve desktop view'larda responsive çalışıyor
- [ ] Unit testler eklendi

omarreis01:
# Description
Fixes #488

Adds a **View Timeline** link next to **Read more** on every map pin popup. Tapping it opens the existing `/timeline` page with proximity params pre-applied (`?latitude=…&longitude=…&radius_km=0.5`) — the same URL shape the proximity filter chip already produces, so the existing `useFilterState` + `TimelinePage` machinery handles it without a new page or layout.

This PR is also the first place in the app where the proximity centre comes from somewhere other than the user's device, so it adds a small layer of UI affordances to make that distinction visible.

## Backend audit (no backend changes)

| Concern | Backend reality | How the PR handles it |
|---|---|---|
| Proximity-supporting endpoint | `/stories/feed/` accepts `latitude`+`longitude`+`radius_km` (validated as a triple, `radius_km ∈ [0.001, 500]`); haversine post-filter gives an exact circle. | Frontend sends `radius_km=0.5` — well within range. |
| Existing service path | `timelineService.getTimeline` already routes proximity-filter requests through the fallback (feed → re-sort by historical year, capped at 100 rows). | Reused unchanged. |

## Frontend changes

**Pin → /timeline navigation**
- `StoryPopup`: extra anchor → `/timeline?latitude=…&longitude=…&radius_km=0.5`. Coords are rounded with `.toFixed(6)` (~10 cm precision) so raw GeoJSON floats don't pollute the URL or break the device-source tolerance check.
- `mapFeatureUtils.featureToStory`: passes coordinates through from the GeoJSON `Point` geometry (RFC 7946 `[lng, lat]`).
- `MapView.StoryLinkInterceptor`: selector is `a[href^='/stories/'], a[href^='/timeline?']` so popup-link clicks stay SPA-side and preserve `from` state for back-nav. Anchored on `'/timeline?'` so unrelated paths like `/timelines/…` aren't swallowed.

**Device-vs-external proximity centre**
- `deviceLocationService` writes the last granted coords to `sessionStorage` on every successful `getCurrentDeviceCoordinates()` call.
- New `isProximityFromDeviceLocation(lat, lng)` helper compares an active centre against that stash with an ~11 m tolerance — covers the 4-decimal display rounding without confusing two real points. Anything that doesn't match is treated as externally supplied.
- `ActiveFilters`: proximity chip reads `Within 500 m of selected location` when the centre isn't from this session's geolocation grant; plain `Within 500 m` when it is. Threaded `latitude`/`longitude` through `SearchFilter` and `TimelinePage` so the chip can run the check.
- `FilterPanel`:
  - Status copy on open with externally-supplied coords reads `Using a location selected on the map.` (instead of `Using your current location.`).
  - The radio-group `aria-label` flips to `Distance from selected location` in the same case.
  - A `Use my location instead` link button appears next to the status. Clicking it discards the external coords, calls `getCurrentDeviceCoordinates()`, and on grant flips status + label back to the device wording. Defaults to `DEFAULT_PROXIMITY_RADIUS_KM` (0.5 km) when no radius was already picked.

**Shared utility**
- `utils/distance.js`: new `formatDistanceKm` shared with `ActiveFilters` (de-duplicated the `< 1 km → metres` expression that was previously inlined).

## Tests

Full unit coverage for the popup link, link interception (positive + negative `/timelines/…` case), the device-vs-external chip wording, the FilterPanel external-coords status / radio aria-label / "Use my location instead" affordance, the click-to-resolve flow, the Apply-after-switch path, plus service-level tests for the sessionStorage write, the tolerance match, and the non-finite-input guards.

**1066 frontend tests pass; lint and build green.**

# Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Documentation

# Screenshots / Recordings

# Estimated Review Time
- [ ] < 5 min
- [x] 5-15 min
- [ ] > 15 min

🤖 Generated with [Claude Code](https://claude.com/claude-code)


sezindogan (changes requested):
## Design Feedback (Primary)

  The current approach needs to change before this can be merged.

  Clicking "View Timeline" from a map pin popup navigates to a new, separate /nearby-timeline route. This is the wrong UX pattern. The existing timeline view should open with the proximity filter pre-applied — 500 m radius centered on the pin's coordinates rather than a new dedicated page.

 ### What this should look like instead:

  - Clicking "View Timeline" on a pin popup navigates to /timeline with latitude, longitude, and radiusKm=0.5 written into the URL params.
  - The existing TimelinePage + useFilterState already reads those params and the filtering machinery is already wired up — the popup just needs to point there.
  - NearbyTimelinePage, SnakeTimelineView, and the /nearby-timeline route should all be removed.
  - The behaviour should match what tapping a proximity filter chip already does (open the timeline view with the filter set), not push to a new route. The mobile implementation could be a good guideline.

  Responding to the "Notes for reviewer" section directly:

1. The chrome is not a problem — it is useful context. A user who lands on the timeline from a map pin should still be able to see and clear the active proximity filter chip, adjust the radius, or add other filters on top. Hiding that chrome in a stripped-down page removes capability without clear benefit. If the chrome feels like too much, that is a design conversation to have separately; it is not a reason to introduce a new page and a parallel data-fetching path.


 2. The snake layout should not be introduced at all as part of this PR. The snake/zigzag layout was mentioned in issue #488 as a visual idea, but it is not a stated acceptance criterion, and it is not something the rest of the team has signed off on as a design direction. The existing vertical timeline is the established pattern. Introducing a new layout in a PR whose core goal is "add a View Timeline link to map pin popups" is scope creep — and it ends up creating a layout that only appears on one entry point, which fragments the timeline UX. If the snake layout is wanted, it should be a separate, explicitly scoped piece of work with proper design review. Do not change the existing TimelinePage or TimelineView as part of this PR.

  This should end up being a much smaller change: URL construction in StoryPopup pointing at the existing /timeline route, plus the coordinate extraction in mapFeatureUtils which is already correct.


 ## Code Review
  
  Assuming the design is reworked, NearbyTimelinePage.jsx, SnakeTimelineView.jsx, and the /nearby-timeline route will be deleted. Notes below focus on what would survive.

 - frontend/src/utils/distance.js — Good extraction. formatDistanceKm is a clean, testable utility and the refactor in ActiveFilters.jsx is correct. Keep this regardless.

-  frontend/src/components/MapView/mapFeatureUtils.jsx — Extracting location_lat/location_lng from GeoJSON coordinates is correct and necessary. The [lng, lat] ordering note is helpful. Keep this.

-  frontend/src/components/MapView/StoryPopup.jsx — The coord extraction and hasCoords guard are fine. The only change needed is nearbyHref pointing to /timeline?latitude=…&longitude=…&radiusKm=0.5 instead of /nearby-timeline.

-  frontend/src/components/MapView/MapView.jsx — The POPUP_LINK_SELECTOR pattern is fine; update it to intercept /timeline links instead of /nearby-timeline.

 - frontend/src/pages/NearbyTimelinePage.jsx — Well-structured internally (stale-response guard via generationRef, hasValidCoords validation, aria-busy/aria-live accessibility). Delete the file, but carry these patterns forward if any new data-fetching behaviour is ever added to TimelinePage.

-  frontend/src/components/Timeline/SnakeTimelineView.jsx — Delete. Not needed for this feature and not approved as a design direction.

 - Tests — Delete tests for NearbyTimelinePage and SnakeTimelineView along with the components. Update the StoryPopup test to reflect the corrected link destination.

 ## Summary:
The utility extraction (distance.js, coord forwarding in mapFeatureUtils) is solid and should be kept. Everything else should be reworked so "View Timeline" links to the existing /timeline page with proximity params pre-set. No new page, no new layout, no changes to the existing timeline.

omarreis01:
Thanks for the thorough review @sezindogan — fully agreed, applied in 087a2a7.

What changed:
- ✅ `StoryPopup` "View Timeline" now links to `/timeline?latitude=…&longitude=…&radius_km=0.5` (existing TimelinePage + useFilterState handle it).
- ✅ `MapView.StoryLinkInterceptor` selector updated to intercept `/timeline` so the popup click stays SPA-side and preserves the `from` state for back-nav.
- ✅ Deleted: `NearbyTimelinePage.jsx`, `SnakeTimelineView.jsx`, their tests, and the `/nearby-timeline` route in `App.jsx`.
- ✅ Kept (per your notes): `utils/distance.js`, the coord forwarding in `mapFeatureUtils.jsx`, the `hasCoords` guard in `StoryPopup`, and the `formatDistanceKm` refactor in `ActiveFilters`.

Verification: lint ✓ · 1048/1048 frontend tests ✓ · build ✓.

Net diff is now what you described — popup URL construction + the small mapFeatureUtils coord extraction + the shared distance util. Ready for another look when you have a moment.

sezindogan:
The diff is exactly the small surface requested: popup URL construction + mapFeatureUtils coord extraction + shared distance util + selector tweak. The core change is approved.

  One optional nitpick: the new MapView.test.jsx assertion checks for &amp;-escaped HTML in the featurePopupHtml path — correct, but a one-line comment would save a future reader from wondering if it's a mistake. 

 ## UX issues to address 
  These are all consequences of this PR being the first place in the app where proximity coords are set from an external source (a map pin) rather than the device.

  1. Proximity chip doesn't indicate the centre is a selected location, not the user

  ActiveFilters.jsx:83 renders Within 500 m regardless of whether the centre is the user's device or a pin they tapped. The panel's radio group is even labelled "Distance from current location" (FilterPanel.jsx:420), which is a lie when the coords came from a popup link.

  Suggested fix — pick one or both:
  - Change the chip to Within 500 m of selected location when the coords were not obtained from geolocation in this session.
  - When the filter panel opens with externally-supplied coords, show Using a location selected on the map. in the status line slot that currently shows Using your current location. (FilterPanel.jsx:52).
  
  2. There is no way to switch to geolocation without resetting the filter

  FilterPanel.jsx:128 only calls getCurrentDeviceCoordinates() when localCoords is null. When the panel opens with pin coords already in the URL, localCoords is pre-initialised (FilterPanel.jsx:46-48), so picking any radius just reuses the pin — there is no in-panel path to switch the centre to the user's own location without first clicking Anywhere and re-picking a radius.

  Suggested fix:
  - When the panel opens with externally-supplied coords and no geolocation grant has happened this session, show a "Use my location instead" button next to the status message. Clicking it clears localCoords and fires getCurrentDeviceCoordinates(), swapping the centre to the device — consistent with the existing radius-change flow. Nothing changes if geolocation is used in the first place.

omarreis01:
Thanks for the follow-up @sezindogan — both rounds applied in 556817a.

**Optional nitpick (done)**
- One-line comment added to the `featurePopupHtml` test explaining that the `&amp;` is `renderToStaticMarkup`'s escape, not a mistake.

**UX issues (done)**

How "device vs external" is detected: `deviceLocationService` now stashes the last granted coords in `sessionStorage` on every successful `getCurrentDeviceCoordinates()` call, and a new `isProximityFromDeviceLocation(lat, lng)` helper compares the active centre against that stash with an ~11 m tolerance (covers the 4-decimal display rounding). Anything that doesn't match → "external" (today, only the map-pin "View Timeline" link).

1. **Proximity chip wording** — `ActiveFilters` now reads `Within 500 m of selected location` when the centre isn't from this session's geolocation grant; plain `Within 500 m` when it is. Threaded `latitude`/`longitude` through `SearchFilter` and `TimelinePage` so the chip can run the check.

2. **FilterPanel status copy** — On open with externally-supplied coords, the status line reads `Using a location selected on the map.` instead of `Using your current location.`. The radiogroup's `aria-label` flips to `Distance from selected location` in the same case.

3. **"Use my location instead"** — A link button appears next to the status whenever the centre is external. Clicking it clears the pin coords, fires `getCurrentDeviceCoordinates()`, and on grant flips the status + label back to device wording. Defaults to a 500 m radius if one wasn't already picked. Apply commits the device coords, not the original pin coords.

**Tests added**: chip wording (both directions), status copy, radio aria-label, affordance presence/absence, click-to-resolve flow, Apply-after-switch path, plus service-level tests for the sessionStorage write, the tolerance match, and the non-finite-input guards. **1064 frontend tests pass; lint and build green.**

Ready for another look.

sezindogan:
Overview: Adds a "View Timeline" link to every map pin popup that navigates to /timeline?latitude=…&longitude=…&radius_km=0.5. Also adds infrastructure to distinguish device-granted coordinates from externally-supplied ones (sessionStorage heuristic), updates the proximity chip/filter copy accordingly, and extracts formatDistanceKm.

##  Issues

  1. POPUP_LINK_SELECTOR is overly broad

  // MapView.jsx:98
  const POPUP_LINK_SELECTOR = "a[href^='/stories/'], a[href^='/timeline']";

  a[href^='/timeline'] matches anything starting with /timeline — so /timelines/…, /timeline-of-events, etc. would also be intercepted. Since the only target is /timeline?…, this should be a[href^='/timeline?'] to be safe.

  2. No coordinate rounding before embedding in the URL

  // StoryPopup.jsx:132-134
  const nearbyHref = hasCoords
  ? `/timeline?latitude=${lat}&longitude=${lng}&radius_km=0.5`
  : null;

  lat and lng are raw floats from GeoJSON, which can be 15+ decimal places (41.012345678901234). This clutters the URL, breaks isProximityFromDeviceLocation's tolerance assumption if the stored device coords were rounded differently, and looks bad in the address bar. Round to 6 decimal places (~10 cm precision, well beyond what's needed here): lat.toFixed(6).

 ## Minor / Non-blocking

  - The sessionStorage-based heuristic in isProximityFromDeviceLocation is a reasonable tradeoff but worth noting: if the user is physically standing within ~11 m of a pin they click, the chip will incorrectly say "Within X" instead of "Within X of selected location". Documented in the code; acceptable for now.
  - handleSwitchToDeviceLocation defaults localRadiusKm to 0.5 when no radius is selected. The comment calls this "atypical" — the test for it is good, and the behaviour is fine, but consider whether 0.5 should be a named constant (DEFAULT_PROXIMITY_RADIUS_KM) shared with the NearbyTimeline usage rather than a magic literal.
  - Test coverage is thorough — all the edge cases for missing/invalid coords and the "Use my location instead" flow are covered. Good.

 ## Summary

  Two actionable fixes before merge: update the PR description to match the actual implementation, and either tighten POPUP_LINK_SELECTOR or round coordinates in the nearbyHref URL. The coord-rounding issue is the most likely to cause a real bug (isProximityFromDeviceLocation tolerance mismatch).

omarreis01:
Thanks again @sezindogan — applied in b42f2c5.

**Actionable**

- ✅ **`POPUP_LINK_SELECTOR` tightened** to `a[href^='/timeline?']`. Sibling paths like `/timelines/…` or `/timeline-of-events` are no longer swallowed by the click interceptor. Added a negative-case test for it.
- ✅ **Coords rounded** to 6 decimals (~10 cm) via `.toFixed(6)` before going into `nearbyHref`. Drops the 15-decimal GeoJSON tail and keeps the comparison in `isProximityFromDeviceLocation` comfortably inside its 1e-4 tolerance window. Updated the `StoryPopup` + `featurePopupHtml` tests to expect the rounded form, and added a case proving high-precision input is truncated correctly.
- ✅ **PR description rewritten** to reflect the actual implementation (the original body still referenced `NearbyTimelinePage` / snake layout / etc.).

**Minor**

- ✅ Extracted `DEFAULT_PROXIMITY_RADIUS_KM = 0.5` in `FilterPanel`, kept in sync with `PROXIMITY_OPTIONS`. The switch-to-device fallback no longer relies on a magic literal.
- The 11 m tolerance edge case (user physically standing within ~11 m of the pin they tap) is acknowledged — keeping the heuristic as documented in the code.

**Verification:** lint ✓ · 1066/1066 frontend tests ✓ · build ✓.

sezindogan (approved):
Looks good. Ready for merge.

PR #600, opened 2026-05-09
Feat/frontend admin panel

### LINKED ISSUE #211: [Feature][Frontend] Admin moderation panel for reports, content, users, and tags
### Description

Implement an admin panel that provides moderation and management capabilities for the platform. This is a protected, admin-only section accessible at /admin, where authorized users can review reports, manage content, control users, and maintain tags.

The panel should include a sidebar navigation with separate pages for Reports, Stories, Users, and Tags. Each page should present data in a clear, paginated table or list, with search and filtering where applicable. Admins should be able to resolve reports with appropriate actions, remove stories with a reason, ban or unban users, and delete tags.

All destructive actions (e.g., removing content, banning users) must require confirmation via modal dialogs. The interface should handle loading states properly and restrict access for non-admin users.

Integrate all functionality through the admin service API and include unit tests covering route protection, core actions, filtering, and confirmation flows.

### Acceptance Criteria

- [ ] Admin panel at `/admin` protected by admin role check
- [ ] Non-admin users redirected or shown "Not authorized"
- [ ] Admin sidebar navigation: Reports, Stories, Users, Tags
- [ ] **Reports page:** table with reporter, target, reason, status, date; filter by status; resolve with outcome (no action / remove content / ban user)
- [ ] **Stories page:** searchable list; remove button with reason input
- [ ] **Users page:** searchable list; ban/unban toggle
- [ ] **Tags page:** list with story count; remove button
- [ ] Confirmation modals before destructive actions (remove, ban)
- [ ] Pagination on all list views
- [ ] Loading states on all sub-pages
- [ ] API service: `adminService` with all CRUD functions
- [ ] Unit tests: route protection, report list/filter/resolve, story removal, user ban/unban, tag removal, confirmation modals

### Meta (optional)

- Effort: L
- Priority: P1
- Week: 4

sezindogan:
# 📝 Description
Fixes #211 
Adds an admin moderation panel at /admin, gated by user.role === "admin". Sidebar navigation routes to four sections (Reports, Stories, Users, Tags), each with paginated lists, filtering/search, and confirmation modals on every destructive action. The "Admin" link in the header only renders for admin users; non-admins hitting /admin see a "Not authorized" page; unauthenticated users get redirected to /login.

  All moderation calls go through a new adminService that wraps the /moderation/* endpoints. Comment-report removal is handled correctly given the cascade on Report.comment — the resolve PATCH is skipped because the cascade already removes the report row.

  Two backend gaps surfaced during the build are listed below as follow-ups; the panel still works without them, just with a degraded Users page.

 ## New features

  - /admin panel with sidebar (Reports, Stories, Users, Tags) and admin-only nav link in the header
  - Reports table with status filter and resolve modal supporting outcomes: no action / remove content / ban user (story author looked up automatically)
  - Stories search + remove with required reason
  - Users lookup (by ID, until a backend list endpoint exists) + ban with confirmation
  - Tags list with story count and remove with confirmation
  - Confirmation modals on every destructive action; pagination on every list view; loading/error/empty states

 ## Modified files

  - src/App.jsx — new admin routes
  - src/components/AppLayout/AppLayout.jsx — role-gated admin link
  - src/services/adminService.js (new) — listReports, resolveReport, removeStory, removeComment, banUser, deleteTag, resolveReportWithAction
  - src/components/AdminRoute.jsx (new) — role-based route guard
  - src/components/AdminPanel/{AdminLayout,Pagination,ConfirmDialog}.jsx (new)
  - src/pages/admin/{AdminReportsPage,AdminStoriesPage,AdminUsersPage,AdminTagsPage}.jsx (new)

 ## Tests

  - adminService.test.js — 11 tests covering all CRUD wrappers and the resolve-with-action flow (including the comment-cascade case)
  - AdminRoute.test.jsx — unauthenticated redirect, non-admin "Not authorized", admin pass-through
  - AdminReportsPage.test.jsx — load, status filter, resolve with note, story-author lookup for ban
  - AdminStoriesPage.test.jsx — load, search, reason-required confirmation, remove flow
  - AdminUsersPage.test.jsx — lookup, 404 handling, ban with confirmation
  - AdminTagsPage.test.jsx — load, search filter, remove confirmation
  - All checks pass: npm run lint, npm run test:run (1070/1070), npm run build

##  Backend follow-ups (not blockers)

  - No /moderation/users/ list endpoint, so the Users page is single-result lookup by ID
  - No unban endpoint, so the ban toggle is one-way
  - GET /users/<id>/ returns 404 for is_active=False, hiding banned users from both the admin lookup and viewers reaching them via story author links

# 📊 Type of Change
- [ ]  Bug fix
- [x]  New feature
- [ ]  Refactoring
- [ ]  Documentation
 
# 📸 Screenshots / Recordings
Screenshot 2026-05-09 at 14 44 15
Screenshot 2026-05-09 at 14 47 45
Screenshot 2026-05-09 at 14 47 40
Screenshot 2026-05-09 at 14 47 34

# ⏱️ Estimated Review Time
- [ ] < 5 min
- [ ] 5-15 min
- [x] > 15 min

### DIFF
--- a/frontend/src/App.jsx
+++ b/frontend/src/App.jsx
@@ -1,7 +1,13 @@
 import { BrowserRouter, Routes, Route } from "react-router-dom";
 
 import ProtectedRoute from "@/components/ProtectedRoute";
+import AdminRoute from "@/components/AdminRoute";
 import AppLayout from "@/components/AppLayout/AppLayout";
+import AdminLayout from "@/components/AdminPanel/AdminLayout";
+import AdminReportsPage from "@/pages/admin/AdminReportsPage";
+import AdminStoriesPage from "@/pages/admin/AdminStoriesPage";
+import AdminUsersPage from "@/pages/admin/AdminUsersPage";
+import AdminTagsPage from "@/pages/admin/AdminTagsPage";
 import { AuthProvider } from "@/context/AuthContext";
 import { ToastProvider } from "@/context/ToastContext";
 import { Toaster } from "@/components/ui/toaster";
@@ -71,6 +77,20 @@ function App() {
               <Route path="/profile/:userId" element={<ProfilePage />} />
               <Route path="/stories/:id" element={<StoryDetailPage />} />
               <Route path="/tags/:slug" element={<TagPage />} />
+              <Route
+                path="/admin"
+                element={(
+                  <AdminRoute>
+                    <AdminLayout />
+                  </AdminRoute>
+                )}
+              >
+                <Route index element={<AdminReportsPage />} />
+                <Route path="reports" element={<AdminReportsPage />} />
+                <Route path="stories" element={<AdminStoriesPage />} />
+                <Route path="users" element={<AdminUsersPage />} />
+                <Route path="tags" element={<AdminTagsPage />} />
+              </Route>
               <Route path="*" element={<NotFoundPage />} />
             </Route>
           </Routes>
--- a/frontend/src/components/AdminPanel/AdminLayout.jsx
+++ b/frontend/src/components/AdminPanel/AdminLayout.jsx
@@ -1,0 +1,52 @@
+import { NavLink, Outlet } from "react-router-dom";
+import { Flag, FileText, Users, Tag as TagIcon, Shield } from "lucide-react";
+
+import { cn } from "@/lib/utils";
+
+const sections = [
+  { to: "/admin/reports", label: "Reports", icon: Flag },
+  { to: "/admin/stories", label: "Stories", icon: FileText },
+  { to: "/admin/users", label: "Users", icon: Users },
+  { to: "/admin/tags", label: "Tags", icon: TagIcon },
+];
+
+function AdminLayout() {
+  return (
+    <div className="container mx-auto max-w-7xl px-4 py-6">
+      <div className="mb-6 flex items-center gap-2">
+        <Shield className="h-6 w-6 text-primary" aria-hidden="true" />
+        <h1 className="text-2xl font-semibold">Admin Panel</h1>
+      </div>
+
+      <div className="grid gap-6 md:grid-cols-[200px_1fr]">
+        <aside aria-label="Admin sections">
+          <nav className="flex flex-row gap-1 overflow-x-auto md:flex-col">
+            {sections.map((s) => (
+              <NavLink
+                key={s.to}
+                to={s.to}
+                className={({ isActive }) =>
+                  cn(
+                    "flex items-center gap-2 rounded-md px-3 py-2 text-sm font-medium transition-colors whitespace-nowrap",
+                    isActive
+                      ? "bg-primary text-primary-foreground"
+                      : "text-muted-foreground hover:bg-muted hover:text-foreground"
+                  )
+                }
+              >
+                <s.icon className="h-4 w-4" aria-hidden="true" />
+                {s.label}
+              </NavLink>
+            ))}
+          </nav>
+        </aside>
+
+        <main className="min-w-0">
+          <Outlet />
+        </main>
+      </div>
+    </div>
+  );
+}
+
+export default AdminLayout;
--- a/frontend/src/components/AdminPanel/ConfirmDialog.jsx
+++ b/frontend/src/components/AdminPanel/ConfirmDialog.jsx
@@ -1,0 +1,57 @@
+import {
+  AlertDialog,
+  AlertDialogAction,
+  AlertDialogCancel,
+  AlertDialogContent,
+  AlertDialogDescription,
+  AlertDialogFooter,
+  AlertDialogHeader,
+  AlertDialogTitle,
+} from "@/components/ui/alert-dialog";
+
+/**
+ * Confirmation dialog for destructive admin actions.
+ * Children are rendered inside the body — used to slot in reason/note inputs.
+ */
+function ConfirmDialog({
+  open,
+  onOpenChange,
+  title,
+  description,
+  confirmLabel = "Confirm",
+  cancelLabel = "Cancel",
+  onConfirm,
+  confirming = false,
+  destructive = true,
+  children,
+  confirmDisabled = false,
+}) {
+  return (
+    <AlertDialog open={open} onOpenChange={onOpenChange}>
+      <AlertDialogContent>
+        <AlertDialogHeader>
+          <AlertDialogTitle>{title}</AlertDialogTitle>
+          {description && (
+            <AlertDialogDescription>{description}</AlertDialogDescription>
+          )}
+        </AlertDialogHeader>
+        {children && <div className="space-y-2">{children}</div>}
+        <AlertDialogFooter>
+          <AlertDialogCancel disabled={confirming}>{cancelLabel}</AlertDialogCancel>
+          <AlertDialogAction
+            variant={destructive ? "destructive" : "default"}
+            disabled={confirming || confirmDisabled}
+            onClick={(e) => {
+              e.preventDefault();
+              onConfirm();
+            }}
+          >
+            {confirming ? "Working…" : confirmLabel}
+          </AlertDialogAction>
+        </AlertDialogFooter>
+      </AlertDialogContent>
+    </AlertDialog>
+  );
+}
+
+export default ConfirmDialog;
--- a/frontend/src/components/AdminPanel/Pagination.jsx
+++ b/frontend/src/components/AdminPanel/Pagination.jsx
@@ -1,0 +1,38 @@
+import { Button } from "@/components/ui/button";
+
+/**
+ * Simple paginator for DRF-style { count, next, previous } responses.
+ * Page numbers are 1-based. Renders nothing when there is only one page.
+ */
+function Pagination({ page, count, pageSize, onPageChange, disabled = false }) {
+  const totalPages = count > 0 ? Math.ceil(count / pageSize) : 1;
+  if (totalPages <= 1) return null;
+
+  return (
+    <div className="mt-4 flex items-center justify-between gap-2">
+      <span className="text-sm text-muted-foreground">
+        Page {page} of {totalPages} ({count} total)
+      </span>
+      <div className="flex gap-2">
+        <Button
+          variant="outline"
+          size="sm"
+          disabled={disabled || page <= 1}
+          onClick={() => onPageChange(page - 1)}
+        >
+          Previous
+        </Button>
+        <Button
+          variant="outline"
+          size="sm"
+          disabled={disabled || page >= totalPages}
+          onClick={() => onPageChange(page + 1)}
+        >
+          Next
+        </Button>
+      </div>
+    </div>
+  );
+}
+
+export default Pagination;
--- a/frontend/src/components/AdminRoute.jsx
+++ b/frontend/src/components/AdminRoute.jsx
@@ -1,0 +1,33 @@
+import { Navigate, useLocation } from "react-router-dom";
+
+import { useAuth } from "@/hooks/useAuth";
+
+/**
+ * Restricts a route to authenticated admin users.
+ *  - Unauthenticated → redirect to /login.
+ *  - Authenticated non-admin → render a simple "Not authorized" page.
+ *  - Admin → render children.
+ */
+function AdminRoute({ children }) {
+  const { user, isAuthenticated } = useAuth();
+  const location = useLocation();
+
+  if (!isAuthenticated) {
+    return <Navigate to="/login" replace state={{ from: location }} />;
+  }
+
+  if (user?.role !== "admin") {
+    return (
+      <div className="container mx-auto max-w-xl px-4 py-16 text-center">
+        <h1 className="text-2xl font-semibold">Not authorized</h1>
+        <p className="mt-2 text-muted-foreground">
+          You do not have permission to access the admin panel.
+        </p>
+      </div>
+    );
+  }
+
+  return children;
+}
+
+export default AdminRoute;
--- a/frontend/src/components/AppLayout/AppLayout.jsx
+++ b/frontend/src/components/AppLayout/AppLayout.jsx
@@ -1,6 +1,6 @@
 import { useState } from "react";
 import { Link, Outlet, useLocation, useNavigate } from "react-router-dom";
-import { Menu, MapPin, LogOut, User, Plus } from "lucide-react";
+import { Menu, MapPin, LogOut, User, Plus, Shield } from "lucide-react";
 
 import { cn } from "@/lib/utils";
 import { Button } from "@/components/ui/button";
@@ -52,8 +52,12 @@ function AppLayout() {
     navigate("/");
   };
 
+  const isAdmin = user?.role === "admin";
+  const baseLinks = isAdmin
+    ? [...publicLinks, { to: "/admin", label: "Admin", icon: Shield }]
+    : publicLinks;
   const isFilterPage = location.pathname === "/" || location.pathname === "/map";
-  const links = publicLinks.map((link) =>
+  const links = baseLinks.map((link) =>
     link.preserveSearch && isFilterPage && location.search
       ? { ...link, basePath: link.to, to: `${link.to}${location.search}` }
       : link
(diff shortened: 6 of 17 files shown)


omarreis01 (commented):
## PR Review: Feat/frontend admin panel

### Summary
Adds a complete `/admin` moderation panel (Reports / Stories / Users / Tags) gated by `user.role === "admin"`, with a clean `adminService` wrapping the six `/moderation/*` endpoints, plus role-aware navigation, route guards, and confirmation modals on every destructive action. Code quality is high, the frontend↔backend contracts match exactly, and test coverage is thorough (1070 tests passing per the PR description). Approving with a few non-blocking notes — almost everything below is informational or a future-scaling concern.

---

### Severity legend
**critical** = block merge · **high** = block merge · **medium** = should fix soon, doesn't block · **low** = nice-to-have / minor · **none** = positive observation

---

### Findings — Frontend

#### 🟡 medium — Admin Stories page hides moderated content
[AdminStoriesPage.jsx](frontend/src/pages/admin/AdminStoriesPage.jsx) calls `getStories({ q, page, pageSize })`, which under the hood hits `/stories/feed/` (or `/stories/search/` when `q` is present). Both endpoints filter to `status=STATUS_PUBLISHED` only ([backend/apps/stories/views.py:186](backend/apps/stories/views.py:186) and [services.py:122](backend/apps/stories/services.py:122)).

Practical effect: once an admin clicks **Remove**, the story status flips to `REMOVED` and the row immediately falls out of the listing — there's no way for the admin to see what they (or another admin) have already moderated, or to coordinate with the backend team about restoring something removed by mistake. This is a UX gap rather than a code defect; flagging because moderation panels usually want a status filter (`Published / Removed / All`). Frontend can't fix this alone — it'd need an admin-scoped list endpoint that doesn't filter by status. Surfacing for awareness, not as a blocker.

#### 🟡 medium — `AdminUsersPage` test mock doesn't match real API response shape
[AdminUsersPage.test.jsx:32](frontend/src/pages/__tests__/AdminUsersPage.test.jsx) mocks `getPublicProfile.mockResolvedValue({ id: 12, username: "carol", is_active: true })`, and the page reads `profile.is_active` to decide whether to show "Active" vs "Banned" and to enable the Ban button.

In production, `/users/<id>/` returns a `PublicUserProfileSerializer` payload that **does not include `is_active`** (or `email`) — see [backend/apps/users/serializers.py:168-237](backend/apps/users/serializers.py:168). The flow still works in production *by coincidence*: `get_public_profile` filters to `is_active=True`, so any successfully looked-up user is active, and `profile.is_active === false` evaluates to `false` (since the field is `undefined`) — the page shows "Active" and the Ban button is enabled, which is the right behaviour. But the unit test passes for the wrong reason, and a future change to either side could mask a regression.

Cheapest fix: update the mock to drop `is_active` and assert the page renders "Active" anyway (which is the real behaviour). Or remove the `is_active` branch from the page and always render "Active" pre-ban (since the lookup endpoint can't return anything else).

The `email` line at [AdminUsersPage.jsx:117](frontend/src/pages/admin/AdminUsersPage.jsx) (`profile.email ? \` · ${profile.email}\` : ""`) is also dead in production — fine, just dead code.

#### 🟡 medium — `AdminTagsPage` client-side pagination over an un-paginated response
[AdminTagsPage.jsx](frontend/src/pages/admin/AdminTagsPage.jsx) calls `searchTags(q)` which returns the full list (the backend `/tags/` endpoint isn't paginated — see [backend/apps/tags/views.py:13](backend/apps/tags/views.py:13)). The page then slices `tags.slice(start, start + PAGE_SIZE)` and feeds `count={tags.length}` to `<Pagination>`. Works fine at current tag volumes, but: (a) every page change triggers a full re-fetch via `useEffect(fetchTags, [fetchTags])` even though the data is already in memory — wasteful round trips; (b) at scale the single response gets large and slow. Two cheap improvements: don't refetch on page change (the `query` dependency is what should drive refetches), and switch to server-side pagination once the backend supports it.

#### 🟢 low — `resolveReportWithAction` is two HTTP calls without a transaction
[adminService.js:56-67](frontend/src/services/adminService.js): for `remove_content` (story) and `ban_user`, the side action and the resolve PATCH are sequential. If the second call fails, the system ends up partially inconsistent (user banned but report still pending, story removed but report still pending). The underlying mutations are idempotent server-side (`ban_user` is a no-op for already-banned users; soft-removed stories take the second remove harmlessly), so retrying is safe — but the UI just shows a generic "Action failed" toast that doesn't tell the admin which step failed. Consider catching each step separately and surfacing a more useful error.

#### 🟢 low — Stale modal state if dialog closed mid-submit
In [AdminReportsPage.jsx:78](frontend/src/pages/admin/AdminReportsPage.jsx) and [AdminStoriesPage.jsx](frontend/src/pages/admin/AdminStoriesPage.jsx), `onOpenChange` clears `activeReport` / `target` immediately. If the admin presses Escape during a slow `submitResolve`, the in-flight request keeps running, then sets `actionError` on a now-closed dialog — the error is silently swallowed. Minor; would need a cancel-on-unmount or `submitting`-aware close guard to fix properly.

#### 🟢 low — Extra round-trip to look up story author for `ban_user`
`AdminReportsPage` does `getStoryById(report.target_id)` purely to extract `story.user.id` for the ban call. Each ban-via-story-report costs an extra hit on the public story endpoint. Eliminating it requires the backend report list to include `story_author_id`, so noting as informational — frontend code itself is fine.

#### 🟢 low — `removeStory` semantics worth a docstring update
[adminService.js:23](frontend/src/services/adminService.js) is named `removeStory` and uses `DELETE`, but the backend [stories/services.py:86-90](backend/apps/stories/services.py:86) is a *soft* delete (sets `status = REMOVED` + `moderation_reason`). The current JSDoc says "DELETE /moderation/stories/<id>/" which is technically correct but doesn't hint at the soft-delete behaviour. A one-line note ("status is set to REMOVED, content preserved") would help future callers reason about cascade effects.

#### 🟢 low — `AdminTagsPage` pagination uses `tags.length` as the total
[AdminTagsPage.jsx:128](frontend/src/pages/admin/AdminTagsPage.jsx) passes `count={tags.length}`. If the backend later paginates `/tags/` and `searchTags` evolves to return `{count, results}`, this passes `results.length` instead of `count` and silently breaks pagination. Defensively read `tags.count ?? tags.length` once a server-paginated path lands.

#### ⚪ none — Auth gating is correct
[AuthContext.jsx:10-13](frontend/src/context/AuthContext.jsx) hydrates user state from localStorage *synchronously* on first render, so [AdminRoute.jsx](frontend/src/components/AdminRoute.jsx) never flashes the "Not authorized" page or redirects to `/login` for an admin who's already signed in. The header link in [AppLayout.jsx:55-57](frontend/src/components/AppLayout/AppLayout.jsx) uses the same `user?.role === "admin"` check, so the two stay consistent. Backend enforces `IsAdminUser` on every endpoint anyway, so client gating is purely UX. ✅

#### ⚪ none — Accessibility is well-considered
- Forms have `role="search"`, every input has an associated `<Label htmlFor=...>`.
- Confirmation modals use `AlertDialog` (which surfaces as `role="alertdialog"`).
- Sidebar nav has `aria-label="Admin sections"`.
- Disabled state on Ban button when already banned, and on Remove confirm until reason is filled in. ✅

#### ⚪ none — Security
- All dynamic values rendered through React text content — no `dangerouslySetInnerHTML`, no innerHTML mutations.
- ID input on `AdminUsersPage` validated as `Number.isInteger(id) && id > 0`.
- Authorization is server-enforced; the client gate is convenience-only and not bypassable for harm.
- No moderation reasons are reflected back into the DOM in a way that could XSS. ✅

---

### Findings — Tests

Coverage is excellent: 11 service tests, page tests for every admin page, and route-guard tests for `AdminRoute`. Comment-cascade behaviour is explicitly covered, including the "skip the resolve PATCH" edge case. The only test concern is the `is_active`-mock divergence noted above (medium #2) — not a coverage gap, a fidelity gap.

---

Per the project's frontend-only review policy, the items below are observations to help the frontend team coordinate, not asks for changes:

- All six `/moderation/*` endpoints exist and are guarded by `IsAdminUser`. Routes are registered in [config/urls.py:18-24](backend/config/urls.py:18) via per-app `admin_urls.py` modules. ✅
- `Report.comment` FK is `on_delete=CASCADE` ([backend/apps/reports/models.py:48-54](backend/apps/reports/models.py:48)), so the frontend's choice to skip the resolve PATCH after deleting a comment (in `resolveReportWithAction`) is correct — the resolve PATCH would 404. ✅
- The PR's "Backend follow-ups" list is accurate: there is no `/moderation/users/` list endpoint, no unban endpoint, and `get_public_profile` filters `is_active=True` ([backend/apps/users/services.py:346](backend/apps/users/services.py:346)) so banned users 404 from the public profile route.
- `StoryDetailView`'s queryset is *not* status-filtered ([backend/apps/stories/views.py:198](backend/apps/stories/views.py:198)), so `getStoryById` from `AdminReportsPage` continues to work even after a story has been soft-removed — good for the ban-via-report flow.
- Backend ban endpoint returns `{ id, email, username, role, is_active }` (UserBanResponseSerializer), which matches what the frontend reads at [adminService.js:35](frontend/src/services/adminService.js).

---

| Frontend call | Backend route | Status |
|---|---|---|
| `GET /moderation/reports/?status=&page=&page_size=` | `AdminReportListView` (IsAdminUser, StoryPagination) | ✅ match |
| `PATCH /moderation/reports/<id>/resolve/` body `{resolution_note}` | `AdminReportResolveView` (IsAdminUser, `ReportResolveSerializer`) | ✅ match |
| `DELETE /moderation/stories/<id>/` body `{moderation_reason}` | `AdminStoryRemovalView` (IsAdminUser, soft-delete) | ✅ match |
| `DELETE /moderation/comments/<id>/` | `AdminCommentRemovalView` (IsAdminUser, hard-delete) | ✅ match |
| `PATCH /moderation/users/<id>/ban/` | `AdminUserBanView` (IsAdminUser, idempotent) | ✅ match |
| `DELETE /moderation/tags/<id>/` | `AdminTagDeleteView` (IsAdminUser) | ✅ match |
| `GET /stories/feed/` (and `/stories/search/`) used by AdminStoriesPage | published-only filter | ⚠️ filters out removed stories — see medium #1 |
| `GET /users/<id>/` used by AdminUsersPage | `is_active=True` only, payload omits `is_active`/`email` | ⚠️ test mock divergence — see medium #2 |
| `GET /tags/?q=` used by AdminTagsPage | un-paginated list | ⚠️ scaling — see medium #3 |
| `GET /stories/<id>/` used by ban-via-report | `StoryDetailView` queryset is not status-filtered | ✅ works post-removal |

---

### Performance

- Pagination on Reports (page_size=20) and Stories (page_size=12) is server-side — fine. ✅
- Tags page does the entire list in a single response and paginates client-side — see medium #3.
- `useCallback` on `fetchReports` / `fetchStories` / `fetchTags` properly stabilises the `useEffect` dependency. ✅
- Per-ban round-trip to fetch `story.user.id` (low #6) is small but adds up if the admin works through a queue.
- AdminTagsPage refetches on **every** page change even though the full list is already in memory — see medium #3.
- No N+1, no unnecessary re-renders, no leaking subscriptions.

---

### Verdict
**APPROVE** ✅

No blocking issues. The frontend code is well-structured, contracts match the backend exactly, security and a11y are considered, and tests are thorough. The medium-severity items (#1 stories status filter, #2 test mock fidelity, #3 tags pagination) are worth a follow-up ticket but don't gate this merge.

#### Suggested follow-ups (non-blocking)
- **Frontend ticket**: fix `AdminUsersPage.test.jsx` mocks to match the real `PublicUserProfileSerializer` shape, or remove the `is_active`-pre-ban branch since the lookup endpoint can never return a banned user (medium #2).
- **Frontend ticket**: in `AdminTagsPage`, drop the per-page refetch and just slice `tags` locally on `setPage` (medium #3 partial fix). Server-side pagination can come later when the backend supports it.
- **Cross-team coordination ticket**: admin Stories panel needs status visibility (Published / Removed / All) — requires a backend admin list endpoint (medium #1).

Nice work overall — moderation flow logic, especially the comment-cascade handling, is thoughtful. 👍

sezindogan:
Thanks for the thorough review! Applied all actionable suggestions:

- Medium 2 — Updated both AdminUsersPage mocks to drop is_active: true/email (matching the real PublicUserProfileSerializer shape); removed the dead profile.email display line from the page.
- Low — stale modal guard — onOpenChange in AdminReportsPage and AdminStoriesPage now checks !submitting before clearing state, so an in-flight request can't silently discard its error into a closed dialog.
 - Low — AdminTagsPage count — <Pagination count> now reads tags.count ?? tags.length so a future server-paginated response won't silently break  pagination.
 - Low — removeStory docstring — Added a note clarifying the soft-delete behaviour (status=REMOVED, content preserved).

Medium 1 (admin stories status filter) and medium 3 (server-side tag pagination) both require backend changes, which are not of high priority as of now.

Ready for review.

omarreis01 (approved):
## Re-review: fixes verified ✅

Pulled the fix commit ([f60f9bb](https://github.com/bounswe/bounswe2026group4/commit/f60f9bbf)) into a fresh worktree and ran the admin test suite — **6 files, 27 tests all green**. Walked each fix against the original finding:

### Verification

#### ✅ medium #2 — Test mock fidelity & dead email line — **fixed correctly**
- `AdminUsersPage.test.jsx` now mocks `{ id, username }` without the bogus `is_active: true`. Both tests still pass for the **right** reason now:
  - "looks up a user…" — `isBanned = profile.is_active === false` evaluates `undefined === false` → `false` → page renders "Active" → assertion passes.
  - "bans a user…" — after `banUser` resolves with `{ is_active: false }`, the spread updates `profile.is_active`, the button flips to "Already banned" — assertion passes.
- `AdminUsersPage.jsx` — dead `profile.email ? ` · ${profile.email}` : ""` line removed. ✅

#### ✅ low — Stale modal guard — **fixed correctly on the two pages I flagged**
[AdminReportsPage.jsx:166](frontend/src/pages/admin/AdminReportsPage.jsx) and [AdminStoriesPage.jsx:146](frontend/src/pages/admin/AdminStoriesPage.jsx) now read `onOpenChange={(open) => { if (!open && !submitting) setActiveReport(null); }}` — Escape / click-outside no longer drops in-flight requests on the floor. ✅

#### ✅ low — Pagination count — **fixed (cosmetic but harmless)**
[AdminTagsPage.jsx:129](frontend/src/pages/admin/AdminTagsPage.jsx) now uses `count={tags.count ?? tags.length}`. Currently a no-op since `searchTags` returns an array, but if the backend ever paginates `/tags/`, the prop won't silently break. (Note: `tags.slice(...)` and `visibleTags.map(...)` would also need updates to fully support the paginated shape — but that's a future concern, not introduced here.) ✅

#### ✅ low — `removeStory` docstring — **fixed**
[adminService.js:21](frontend/src/services/adminService.js) now reads: *"Soft-delete: sets status=REMOVED, content is preserved server-side."* ✅

#### ⏭️ medium #1 (admin stories status filter) and medium #3 (server-side tag pagination) — **deferred, correctly scoped**
Both genuinely require a backend list endpoint that doesn't exist yet. The decision to defer is reasonable; these are coordination tickets, not frontend bugs.

---

### New observations from this re-review

#### 🟢 low — Modal-guard fix not applied uniformly (incomplete coverage, not a regression)
The same `onOpenChange` pattern that was fixed on Reports/Stories still exists unguarded on the other two admin pages — the race condition I originally described applies to them too. Not introduced by this PR (it was always there), but worth a one-line follow-up:

- [AdminTagsPage.jsx:137](frontend/src/pages/admin/AdminTagsPage.jsx) — `onOpenChange={(open) => { if (!open) setTarget(null); }}` → could become `if (!open && !submitting)`.
- [AdminUsersPage.jsx:126](frontend/src/pages/admin/AdminUsersPage.jsx) — `onOpenChange={setConfirming}` directly forwards to setState; could be wrapped to block while `submitting` is true.

Practical impact: low. Tag deletion and user ban are single network calls and resolve quickly, so the window for the race is small. Flagging only because the fix on the other two pages set a precedent and these two missed it. Not blocking.

---

### Verdict

**APPROVE** ✅

All actionable findings from the original review have been addressed correctly, no regressions introduced, all admin tests still pass (27/27 in the suites I re-ran). Backend-dependent items are correctly deferred to follow-up tickets. Ship it. 🚀

Nice, surgical fix — minimal diff, exactly what each finding called for.

PR #604, opened 2026-05-09
fix: include location_name in story timeline response

### LINKED ISSUE #603: [Feature][Backend] StoryTimelineSerializer should include location_name
### Description

The StoryTimelineSerializer (used by /stories/timeline/) returns location_lat and location_lng but omits location_name. As a result, timeline cards always display raw coordinates (Mapped at 41.0100, 28.9700) instead of the human-readable place name, and tag chips never appear on timeline cards — even though TimelineEntry already has the rendering logic for both fields and both fields exist on the model.

StoryFeedSerializer already exposes location_name and tags. StoryTimelineSerializer should be brought to parity for these two fields so timeline cards render consistently regardless of which serializer path produced the data.

### Acceptance Criteria

- [ ]  StoryTimelineSerializer includes location_name in its fields list.
- [ ] StoryTimelineSerializer includes tags (serialized via TagSerializer(many=True, read_only=True)) in its fields list.
- [ ] GET /stories/timeline/ response includes location_name and tags on each story object.
- [ ] Timeline cards on /timeline display the location name when location_name is set, and fall back to coordinates only when it is not.
- [ ] Timeline cards on /timeline display tag chips when tags are present.
- [ ] Existing timeline endpoint tests updated; new tests added for the two fields.

### Meta (optional)

Effort: S
Priority: P2

AysuKeskin:
# 📝 Description

Fixes #603

This PR updates the story timeline response to include `location_name` in addition to latitude and longitude.

This change was made based on an additional request from the frontend side. The timeline cards already returned location coordinates, but the frontend also needs the human-readable location name to display timeline stories more clearly in the UI.

Updated changes:
- Added `location_name` to `StoryTimelineSerializer`
- Updated the serializer documentation to reflect that timeline cards now include the location name
- Updated the timeline response test to verify that `location_name` is included in the response

# 📊 Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactoring
- [ ] Documentation

- [x] Code follows the project style guidelines
- [x] Self-review performed
- [x] Tests added/updated
- [x] Timeline response now includes `location_name`
- [x] Feed-specific fields are still excluded from the timeline response
- [x] Tests pass locally

# ⏱️ Estimated Review Time

- [x] < 5 min
- [ ] 5–15 min
- [ ] > 15 min

### DIFF
--- a/backend/apps/stories/serializers.py
+++ b/backend/apps/stories/serializers.py
@@ -467,10 +467,10 @@ class StoryTimelineSerializer(serializers.ModelSerializer):
     Minimal read-only serializer for timeline cards.
 
     Returns only the fields needed to render a story on the timeline: identity,
-    title, all time fields (including EDTF temporal_coverage), coordinates, and
-    a representative photo URL. Feed-specific fields (preview_text,
-    contributor_name, like/save state, status, submitted_at) are intentionally
-    excluded.
+    title, all time fields (including EDTF temporal_coverage), coordinates,
+    location name, and a representative photo URL. Feed-specific fields
+    (preview_text, contributor_name, like/save state, status, submitted_at)
+    are intentionally excluded.
     """
 
     photo_url = serializers.SerializerMethodField()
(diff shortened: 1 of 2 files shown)


KmalM (approved):
Clean, minimal fix. `location_name` is a required non-nullable field so it's safe to add directly to the serializer fields list. Test coverage updated correctly. CI passes. Approving.

PR #619, opened 2026-05-10
Fix/ timeline filtering for BC dates

### LINKED ISSUE #617: [Bug][Backend] Fix timeline filtering failure for BC year boundaries in exact-date stories
### Description

There is a backend bug in the timeline filtering logic when year-based filters are used together with `TIME_DATE` stories and **negative year values** (BC dates).

The issue comes from the timeline service logic using `date_value__year__gte` / `date_value__year__lte` for exact-date stories. In MySQL, this translates to `YEAR(date_value)`, and MySQL’s `YEAR()` function does not accept negative year comparisons. As a result, timeline requests with negative `year_from` or `year_to` values can fail with a **500 error** instead of returning a valid filtered result. 

This appears to be specific to the timeline filtering implementation. The feed behavior does not fail in the same way because it does not rely on the same `date_value__year` pattern for this case. 

The expected behavior is that timeline filtering should handle BC year boundaries safely and should not crash when negative year ranges are requested. Since `date_value` itself cannot meaningfully represent BC dates in this path, the filtering logic for `TIME_DATE` stories should avoid applying MySQL year extraction when the boundary is negative, while still allowing the other supported time types to be filtered correctly. 

This issue is about fixing the backend timeline filter logic so that:

* negative year filters do not produce a server error,
* exact-date stories are handled safely,
* and timeline results remain correct for the supported temporal model.

### Expected Behavior

* Timeline requests with negative `year_from` or `year_to` values do not return 500 errors
* Filtering remains correct for supported story time types
* Exact-date filtering logic does not trigger invalid MySQL `YEAR()` comparisons for BC boundaries

### Acceptance Criteria

* [ ] Timeline filtering no longer returns 500 errors for negative `year_from` or `year_to` values
* [ ] `TIME_DATE` filtering logic is adjusted to avoid invalid MySQL `YEAR()` usage on negative boundaries
* [ ] Other story time types continue to be filtered correctly
* [ ] Existing timeline behavior is preserved for non-negative year filters
* [ ] Tests are updated for negative-year timeline filtering cases if needed
* [ ] Existing relevant tests pass

### Meta (optional)

Estimated Duration: 1–2 hours

AysuKeskin:
# 📝 Description

Fixes #617 

This PR fixes a timeline filtering bug for `exact_date` stories when the requested year boundary is in BC / negative years.

Previously, `date_value__year__gte` and `date_value__year__lte` filters were always applied for `exact_date` timeline entries. However, MySQL `YEAR()` extraction only works reliably for non-negative calendar years. When `year_from` or `year_to` was negative, applying this filter could cause incorrect behavior or prevent filtering from working as expected.

With this change:

- `exact_date` filtering through `date_value__year` is only applied when the boundary year is non-negative.
- For BC boundaries, the MySQL `YEAR()` filter is skipped safely.
- Existing filtering behavior for decade, exact year, approximate year, and year range stories is preserved.

---

# 📊 Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactoring
- [ ] Documentation

---

- [x] Timeline filtering does not apply MySQL `YEAR()` extraction for BC / negative year boundaries.
- [x] `year_from` filtering still works for decade, exact year, approximate year, and year range stories.
- [x] `year_to` filtering still works for supported timeline story types.
- [x] `exact_date` filtering is preserved for non-negative years.
- [x] Existing timeline behavior is not changed outside this edge case.
- [x] Code follows the existing service-layer structure in `stories/services.py`.
- [x] I have self-reviewed the changes.

---

# ⏱️ Estimated Review Time

- [x] <5 minutes
- [ ] 5–15 minutes
- [ ] >15 minutes

### DIFF
--- a/backend/apps/stories/services.py
+++ b/backend/apps/stories/services.py
@@ -188,6 +188,11 @@ def get_story_timeline(
     qs = Story.objects.filter(status=Story.STATUS_PUBLISHED)
 
     if year_from is not None:
+        date_q = (
+            # MySQL YEAR() only handles non-negative years; skip for BC boundaries
+            Q(time_type=Story.TIME_DATE, date_value__year__gte=year_from)
+            if year_from >= 0 else Q()
+        )
         qs = qs.filter(
             # decade: interval ends at year+9; include if year+9 >= year_from ↔ year >= year_from-9
             Q(time_type=Story.TIME_DECADE, year__gte=year_from - 9) |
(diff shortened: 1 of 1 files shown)


KmalM (approved):
Logic is correct and well-reasoned. The asymmetry between the `year_from` and `year_to` BC paths is intentional and sound — MySQL `DATE` fields are always positive years, so including all `TIME_DATE` stories when `year_from < 0` and excluding them entirely when `year_to < 0` is the right call. `Q()` as a no-op in the OR chain is also safe.

One minor note: there are no BC-specific tests added for the timeline endpoint (e.g. `year_from=-500` or `year_to=-100` with a `TIME_DATE` story). The fix is verifiable by inspection given the DB-level constraint, but a regression test would pin this behaviour. Not a blocker.

PR #636, opened 2026-05-10
fix(frontend): remove redundant Back button from Timeline

### LINKED ISSUE #624: [Bug][Frontend] Remove Redundant Back Button from Timeline Filter
### Description

Currently, there is a back button on the Timeline view only when navigated from a 'View Timeline' link. Following a design review, that button on the timeline view has been deemed unnecessary. Since the timeline serves as a filter state rather than a separate page, users should navigate by switching between the map and feed interfaces, which preserves the current filter status. It also causes confusion that the button is visible depending on how timeline was opened.

This change also aligns the desktop web experience with the existing mobile UI, which does not include this button.

### Acceptance Criteria

- [ ] UI Removal: The "Back" button is completely removed from the Timeline component/view.

- [ ] Navigation Consistency: Clicking the 'View Timeline' link on story preview on the map still triggers the timeline filter state without the redundant button appearing.

- [ ] Existing unit tests are updated to reflect the change.

### Meta (optional)

_No response_

omarreis01:
# Description
Fixes #624

Removes the conditional "Back to map" button from the Timeline view per design review. The timeline is a filter state rather than a separate page — users should navigate via the map/feed switcher, which already preserves the active filter. This also aligns the desktop UI with the existing mobile UI, which doesn't render this button.

Changes:
- `frontend/src/pages/TimelinePage.jsx`: removed the `mapBack` button and the now-unused `useLocation`, `useNavigate`, and `ArrowLeft` imports.
- `frontend/src/pages/__tests__/TimelinePage.test.jsx`: inverted the existing back-button tests to assert the button is never rendered (including when navigated from `/map`).

The "View Timeline" link in `StoryPopup` still works — only the button on the receiving end is gone; proximity filter params are still applied as before.

# Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Refactoring
- [ ] Documentation

# Screenshots / Recordings

# Estimated Review Time
- [x] < 5 min
- [ ] 5-15 min
- [ ] > 15 min

### DIFF
--- a/frontend/src/pages/TimelinePage.jsx
+++ b/frontend/src/pages/TimelinePage.jsx
@@ -1,6 +1,4 @@
 import { useCallback, useEffect, useMemo, useRef, useState } from "react";
-import { useLocation, useNavigate } from "react-router-dom";
-import { ArrowLeft } from "lucide-react";
 
 import { Button } from "@/components/ui/button";
 import { ErrorState } from "@/components/ui/error-state";
@@ -42,11 +40,6 @@ function TimelinePage() {
     removeTag,
     clearAll,
   } = useFilterState();
-
-  const routerLocation = useLocation();
-  const navigate = useNavigate();
-  const _from = routerLocation.state?.from;
-  const mapBack = _from === "/map" || _from?.startsWith("/map?") ? _from : null;
 
   const [stories, setStories] = useState([]);
   const [count, setCount] = useState(0);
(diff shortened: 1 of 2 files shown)


sezindogan (approved):
## Overview

Removes the conditional "Back to map" button and its associated useLocation/useNavigate/ArrowLeft dependencies from TimelinePage. Tests are updated to assert the button is never rendered. Clean, minimal, scoped change.

##  Minor Observations (non-blocking)

  - Test coverage consolidation: The three original back-button tests are collapsed into two. The third case (navigated from a non-map page) is dropped entirely. Since the button no longer exists under any condition, this is fine — but one parameterized test asserting not.toBeInTheDocument() for all three scenarios would be slightly more idiomatic. Not worth a change request.
- Leftover Button import: Button from @/components/ui/button is still imported at line 5. If it's used elsewhere in the component (e.g., filter panel or load-more), no action needed — but worth a quick check that it's not now unused.
 - Dead _from variable pattern: The original code used a _from naming convention to signal intentional non-use. It's fully removed here, which is correct — just noting it was a code smell in the original that's been cleaned up.


##  Verdict

Approve. The change is correct, the description is accurate, and the tests are updated appropriately. The PR is straightforward and correct. No logic bugs, no security concerns, no performance implications. Ready for merge.

PR #638, opened 2026-05-11
fix: move delete account/reset-password to profile page and fix wrong…

### LINKED ISSUE #637: [Bug][Frontend] Fix wrong password UI in account deletion
### Description

There is a critical workflow error in the "Delete Profile" process. When a user attempts to delete their account but enters an incorrect password, the system fails to provide validation feedback. Instead, it triggers a success notification ("Profile updated successfully") and closes the modal, even though no action was taken. Additionally, the success messaging used for actual deletions is inconsistent with the action performed.
On a final note, Delete Profile is in Edit Profile section which differs from mobile. 'Delete Profile' and 'Reset Password' buttons should be added in the 'Profile Page', with the' Delete Profile' button removed from 'Edit Profile'.

### Steps to Reproduce

- Log into the application.
- Navigate to User Settings or the Profile page.
- Click the "Delete Account" button to open the confirmation modal.
- Enter an incorrect password into the confirmation text field.
- Click the "Confirm Delete" button.

### Current Behavior

- Validation: The modal closes immediately regardless of password accuracy.
- Feedback: A green toast notification appears stating: "Profile updated successfully."
- Account Status: The account remains active (not deleted), which is correct for a wrong password, but the UI suggests a successful modification occurred.
- On successful account deletion, first the profile updated successfully, then the account deleted toast shows.

### Expected Behavior

- Validation Error: If the password is wrong, the modal must stay open.
- Inline Feedback: An error message should appear directly under the password text box: "Wrong password, please try again."
- Success Feedback: If the password is correct and the account is deleted, the toast notification should read: "Account deleted successfully" (instead of "Profile updated").
 - On successful account deletion, only the account deleted successfully toast shows.

### Acceptance Criteria

- [ ] The confirmation modal must remain open if the password validation fails.
- [ ] An error message stating "Wrong password, please try again" must appear directly beneath the password input field when an incorrect password is submitted.
- [ ] The "Profile updated successfully" toast must be suppressed/prevented from appearing during any part of the deletion workflow.
- [ ] Upon successful account deletion, the system must display a toast notification stating: "Account deleted successfully."
- [ ] The UI must correctly catch the error response from the backend and prevent the success redirect or closure logic from firing.

### Meta (optional)

_No response_

sezindogan:
…-password error handling

# 📝 Description
Fixes #637 

Fixes critical bugs in the Delete Profile workflow and restructures account management actions to match mobile UX. When a user entered an incorrect password during account deletion, the modal closed and a misleading "Profile updated successfully" toast appeared. This was caused by the DeleteAccountDialog being nested inside the EditProfileForm, creating an implicit coupling between the two forms. Moving the dialog to ProfilePage decouples them entirely. Error messaging for wrong-password and success responses has also been corrected.

###  New features

  - Delete Profile button added directly to ProfilePage (visible only on own profile), replacing the one hidden in the Edit Profile form's danger zone
  - Reset Password button added to ProfilePage (visible only on own profile), linking to /forgot-password

###  Modified files

  - src/components/Profile/DeleteAccountDialog.jsx — success toast changed to "Account deleted successfully."; wrong-password detection now triggers on any HTTP 400, suppressing the raw Axios error message
  - src/components/Profile/EditProfileForm.jsx — removed "Danger zone" section (Delete account button and DeleteAccountDialog)
  - src/pages/ProfilePage.jsx — added Delete Profile button, Reset Password link, deleteDialogOpen state, and DeleteAccountDialog mount at page level
  - src/components/Profile/__tests__/DeleteAccountDialog.test.jsx — updated toast assertion; updated wrong-password error text assertions; added regression test for 400 with no password array in body
  - src/components/Profile/__tests__/EditProfileForm.test.jsx — removed danger zone tests and unused mocks
  - src/pages/__tests__/ProfilePage.test.jsx — added DeleteAccountDialog mock; added 6 new tests for Delete Profile and Reset Password buttons

 ### Tests

  - All 1136 existing tests pass
  - New: Delete Profile button shows/hides on own vs. other profile
  - New: clicking Delete Profile opens DeleteAccountDialog; closing it works correctly
  - New: Reset Password link shows/hides on own vs. other profile and points to /forgot-password
  - New: 400 response with no password array in body now shows "Wrong password, please try again." instead of raw Axios message

# 📊 Type of Change
- [x]  Bug fix
- [ ]  New feature
- [ ]  Refactoring
- [ ]  Documentation
 
# 📸 Screenshots / Recordings

# ⏱️ Estimated Review Time
- [x] < 5 min
- [ ] 5-15 min
- [ ] > 15 min

### DIFF
--- a/frontend/src/components/Profile/DeleteAccountDialog.jsx
+++ b/frontend/src/components/Profile/DeleteAccountDialog.jsx
@@ -48,12 +48,16 @@ function DeleteAccountForm({ onOpenChange }) {
     } catch (err) {
       const status = err?.response?.status;
       const data = err?.response?.data;
-      const isPasswordError =
-        status === 400 ||
-        (Array.isArray(data?.password) && data.password.length > 0);
-      const message = isPasswordError
-        ? "Wrong password, please try again."
-        : (data?.detail || err?.message || "Failed to delete account. Please try again.");
+      const passwordFieldError =
+        Array.isArray(data?.password) && data.password.length > 0
+          ? data.password[0]
+          : null;
+      const message =
+        passwordFieldError
+          ?? data?.detail
+          ?? (status === 400 ? "Wrong password, please try again." : null)
+          ?? err?.message
+          ?? "Failed to delete account. Please try again.";
       setError(message);
       setSubmitting(false);
     }
--- a/frontend/src/components/Profile/DeleteAccountDialog.jsx
+++ b/frontend/src/components/Profile/DeleteAccountDialog.jsx
@@ -43,15 +43,17 @@ function DeleteAccountForm({ onOpenChange }) {
       // AuthContext listens for to drop the in-memory user.
       clearTokens();
       onOpenChange(false);
-      toast.success("Your account has been deleted.");
+      toast.success("Account deleted successfully.");
       navigate("/");
     } catch (err) {
+      const status = err?.response?.status;
       const data = err?.response?.data;
-      const message =
-        (Array.isArray(data?.password) && data.password[0]) ||
-        data?.detail ||
-        err?.message ||
-        "Failed to delete account. Please try again.";
+      const isPasswordError =
+        status === 400 ||
+        (Array.isArray(data?.password) && data.password.length > 0);
+      const message = isPasswordError
+        ? "Wrong password, please try again."
+        : (data?.detail || err?.message || "Failed to delete account. Please try again.");
       setError(message);
       setSubmitting(false);
     }
--- a/frontend/src/components/Profile/EditProfileForm.jsx
+++ b/frontend/src/components/Profile/EditProfileForm.jsx
@@ -13,7 +13,6 @@ import {
 import { useAuth } from "@/hooks/useAuth";
 import PhotoUpload from "@/components/Profile/PhotoUpload";
 import VisibilityToggle from "@/components/Profile/VisibilityToggle";
-import DeleteAccountDialog from "@/components/Profile/DeleteAccountDialog";
 
 const MAX_BIO_LENGTH = 500;
 
@@ -24,7 +23,6 @@ function EditProfileForm({ initialProfile, onSave, onCancel }) {
 
   const [saving, setSaving] = useState(false);
   const [errors, setErrors] = useState({});
-  const [deleteDialogOpen, setDeleteDialogOpen] = useState(false);
 
   const [username, setUsername] = useState(() => initialProfile?.username ?? "");
   const [firstName, setFirstName] = useState(() => prof.first_name ?? "");
@@ -297,29 +295,6 @@ function EditProfileForm({ initialProfile, onSave, onCancel }) {
         </Button>
       </div>
 
-      {/* Danger zone — visually de-emphasised so it isn't an accidental tap target. */}
-      <div className="mt-4 border-t pt-6">
-        <h3 className="text-sm font-medium">Danger zone</h3>
-        <div className="mt-2 flex flex-col items-start justify-between gap-2 sm:flex-row sm:items-center">
-          <p className="text-xs text-muted-foreground">
-            Permanently delete your account, stories, comments, and uploads.
-          </p>
-          <Button
-            type="button"
-            variant="outline"
-            size="sm"
-            onClick={() => setDeleteDialogOpen(true)}
-            disabled={saving}
-          >
-            Delete account
-          </Button>
-        </div>
-      </div>
-
-      <DeleteAccountDialog
-        open={deleteDialogOpen}
-        onOpenChange={setDeleteDialogOpen}
-      />
     </form>
   );
 }
--- a/frontend/src/components/Profile/__tests__/DeleteAccountDialog.test.jsx
+++ b/frontend/src/components/Profile/__tests__/DeleteAccountDialog.test.jsx
@@ -96,11 +96,11 @@ describe("DeleteAccountDialog", () => {
 
     await waitFor(() => expect(clearTokens).toHaveBeenCalledTimes(1));
     expect(onOpenChange).toHaveBeenCalledWith(false);
-    expect(successToast).toHaveBeenCalledWith("Your account has been deleted.");
+    expect(successToast).toHaveBeenCalledWith("Account deleted successfully.");
     expect(navigateMock).toHaveBeenCalledWith("/");
   });
 
-  it("surfaces a backend 'incorrect password' error inline and does not navigate", async () => {
+  it("surfaces an inline 'wrong password' error and does not navigate", async () => {
     deleteAccount.mockRejectedValue({
       response: { status: 400, data: { password: ["Incorrect password."] } },
     });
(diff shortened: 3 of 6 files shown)


omarreis01 (commented):
## PR Review: fix — move delete account / reset-password to profile page and fix wrong-password error handling

### Summary
The architectural change — pulling `DeleteAccountDialog` out of `EditProfileForm` and mounting it directly on `ProfilePage` — is the right fix for the reported bug. The implicit form coupling that produced the misleading "Profile updated successfully" toast on wrong-password is fully eliminated by this decoupling. Test coverage for the new buttons and the regression case is solid. There is one notable concern in the new error-handling logic and a few smaller UX/test gaps worth surfacing.

---

### Findings by severity

#### 🚫 Critical
_None._

#### 🔴 High
_None._

#### 🟠 Medium

**M1. Overly aggressive `status === 400 → "Wrong password"` mapping in `DeleteAccountDialog.jsx`.**
The new logic in `frontend/src/components/Profile/DeleteAccountDialog.jsx:46-55` is:

```js
const isPasswordError =
  status === 400 ||
  (Array.isArray(data?.password) && data.password.length > 0);
const message = isPasswordError
  ? "Wrong password, please try again."
  : (data?.detail || err?.message || "Failed to delete account. Please try again.");
```

Treating **any** HTTP 400 as a wrong-password error is broader than necessary. Today the backend's `DeleteAccountSerializer` only has a non-trivial validator on `password`, so 400 ≈ wrong password in practice. But:

- If the serializer ever gains another validated field (e.g. an MFA code, a captcha token, a future "reason" field), a legitimate validation error from that field will be **silently mis-labeled** as a wrong password.
- A malformed `refresh` / `hard_delete` value would also produce 400 with a non-`password` error body and be reported as wrong password.
- The branch order also throws away a more specific message: if the backend responds with `400 { "password": ["Incorrect password."] }`, the user no longer sees "Incorrect password." — they see "Wrong password, please try again." That's a deliberate UX choice (it's friendlier), just flagging that the per-field message is now unreachable.

**Suggested safer pattern:** prefer specific info, fall back to the generic 400 message:

```js
const passwordFieldError =
  Array.isArray(data?.password) && data.password.length > 0
    ? data.password[0]
    : null;
const message =
  passwordFieldError
    ?? data?.detail
    ?? (status === 400 ? "Wrong password, please try again." : null)
    ?? "Failed to delete account. Please try again.";
```

This preserves the bug fix (the raw Axios "Request failed with status code 400" message is suppressed) but doesn't swallow future or unrelated 400s.

**M2. "Reset Password" routes a *logged-in* user to `/forgot-password`.**
On `frontend/src/pages/ProfilePage.jsx:291-296`, the new button links to the unauthenticated forgot-password flow. That means an authenticated user is sent to a page that typically asks for their email and sends a reset link — a strange round-trip for someone who is already authenticated. Semantically "Reset Password" and "Forgot Password" are different products: the former usually asks for *current* password + new password inline. Worth confirming with design/mobile whether the intent is genuinely to re-use the forgot-password flow, or whether a dedicated `/account/change-password` route is planned. If the forgot-password reuse is intentional, consider renaming the button to "Send password reset email" to set expectations.

#### 🟡 Low

**L1. Delete Profile button uses `variant="outline"` rather than `variant="destructive"`.**
The button at `frontend/src/pages/ProfilePage.jsx:297-304` is visually indistinguishable from Notifications / Edit Profile / Reset Password. The previous "Danger zone" section visually de-emphasised the action — that affordance is now gone. The destructive confirm modal still gates the action, so this is not a safety problem, just a visual signaling question. Confirm this matches the mobile UX referenced in the PR description.

**L2. Missing test: buttons should not appear in edit mode.**
The header (and therefore the new buttons) is replaced by `EditProfileForm` while `editMode === true`. There's no explicit assertion of this in `ProfilePage.test.jsx`. Cheap to add and prevents a future regression where someone hoists the buttons out of the conditional.

**L3. Missing test: non-password 400 response shape.**
The new regression test covers `400` with `data: {}` (raw Axios message suppressed), and the existing test covers `400 { password: [...] }`. But there's no test for `400 { detail: "…" }` or `400 { hard_delete: [...] }`. Given M1, asserting the current behavior would at least make the broad mapping explicit and pin it down for future readers.

**L4. `flex-wrap` added without a visual check.**
`ProfilePage.jsx:272` switches the action row to `flex-wrap`. With four buttons (Notifications, Edit Profile, Reset Password, Delete Profile) plus username/avatar on the left, narrow viewports may now wrap the Delete button onto its own line in an unusual position. Worth a quick visual check at ~375px width.

#### ⚪ None / Nits
- The decision to mount `DeleteAccountDialog` at the page level guarded by `isOwnProfile` (`ProfilePage.jsx:373-379`) is a nice defensive touch — even if `setDeleteDialogOpen(true)` were somehow triggered for a non-owner, the dialog won't render.
- Toast string change ("Your account has been deleted." → "Account deleted successfully.") and the corresponding test update are fine; aligns with the PR's stated mobile-parity goal.
- The `DeleteAccountForm` component pattern (separating body from the AlertDialog so Radix remounts it on each open) is preserved correctly — no regression here.
- Cleanup of the now-unused `deleteAccount` mock and `DeleteAccountDialog` mock from `EditProfileForm.test.jsx` is correct.

---

This PR doesn't change any API calls — it only restructures UI and reshapes the local error string. Verified for completeness:

- `DELETE /users/me/` — ✅ endpoint exists (`backend/apps/users/views.py:154`), expects `{password, hard_delete, refresh}` via `DeleteAccountSerializer` (`serializers.py:240-252`). Returns `204` on success and `400 {"password": ["Incorrect password."]}` on wrong password. Frontend's `deleteAccount` in `services/userService.js:93` matches. No contract change required for this PR.

---

### Verdict
**APPROVE with non-blocking comments.**

The PR is well-scoped, correctly fixes the original bug (wrong-password no longer shows a success toast), and adds meaningful tests. No blockers.

**Recommended (non-blocking) follow-ups:**
1. Tighten the 400 → "Wrong password" mapping (M1) so future serializer changes don't get silently mis-labeled.
2. Confirm the "Reset Password" → `/forgot-password` routing is the agreed UX (M2), or rename the button to match what it actually does.
3. Add the small test gaps (L2, L3) when convenient.

Nice work on the architectural cleanup — decoupling the dialog from `EditProfileForm` is the right call and removes a class of latent bugs, not just the reported one.

sezindogan:
Thanks for the thorough review!
  
  M1 — Applied. Replaced the broad status === 400 → "Wrong password" guard with the safer precedence chain you suggested:

  ```
const passwordFieldError =
    Array.isArray(data?.password) && data.password.length > 0
        ? data.password[0]
         : null;
const message =
    passwordFieldError
        ?? data?.detail
        ?? (status === 400 ? "Wrong password, please try again." : null)
        ?? err?.message
        ?? "Failed to delete account. Please try again.";
```

  This surfaces the specific backend field error first (data.password[0]), then detail, then the generic 400 fallback — so future serializer fields won't be silently mislabeled. Updated the two affected tests to match the new behavior.

  M2 — Intentional, same as mobile. Checked the mobile implementation: ProfileScreen calls authService.forgotPassword directly from the profile page (same forgot-password endpoint/flow). The frontend navigating to /forgot-password achieves the same result with a slightly different UX;  this is consistent with the mobile design intent.

  L1 — Applied. Changed Delete Profile button to variant="destructive" to restore the visual danger signal that was previously conveyed by the old "Danger zone" section.

  L2 — Applied. Added a test asserting that all four action buttons/links (Edit Profile, Delete Profile, Reset Password, Notifications) are absent while editMode === true.

  L3 — Applied. Added a test for 400 { detail: "..." } to pin the new behavior: a detail field now surfaces its message rather than being overwritten by the generic 400 string.

  L4 — Noted. Did a quick check at 375px — the four buttons wrap cleanly onto two lines with Delete Profile landing in a reasonable position. No visual regression.

omarreis01 (approved):
Re-reviewed commit `c3a6cb6` ("fix: apply PR review suggestions"). All follow-up items from the previous review are addressed:

- **M1** ✅ — `DeleteAccountDialog.jsx` now uses the safer precedence chain (`data.password[0]` → `data.detail` → 400 fallback → `err.message` → generic). The two existing tests are updated to assert the specific backend field message, and a new test pins the `data.detail` path.
- **M2** ✅ — Confirmed as intentional parity with the mobile `authService.forgotPassword` flow. Fine to keep `/forgot-password` for now; if the team later decides on a logged-in change-password flow, that's a separate piece of work.
- **L1** ✅ — Delete Profile button switched to `variant="destructive"`, restoring the visual danger signal.
- **L2** ✅ — New test "hides action buttons while in edit mode" asserts Notifications / Edit Profile / Reset Password / Delete Profile are all absent during `editMode`.
- **L3** ✅ — New test covers `400 { detail: "..." }`, locking in the new precedence behavior.
- **L4** ✅ — Visual check confirmed at 375px by the author.

CI is green across Backend Tests, Frontend Checks, Frontend E2E, and Mobile Checks.

Approving — good to merge. Thanks for the fast turnaround and the clean follow-up commit. ✅

PR #647, opened 2026-05-12
Feat/docker vite site url build arg

### LINKED ISSUE #645: [DevOps][Frontend] Document frontend environment variables in .env.example
### Description

Currently, the frontend folder lacks a standardized .env.example file. This makes it difficult for new contributors or team members to set up their local development environments, as required environment variables are not explicitly documented. This is also requested by the TAs for easy test setup.

Additionally, a frontend environment variable is not set-up in the production environment and is only working because of a hardcoded fallback. If our code gets deployed on a different domain, it will cause issues.

To improve the developer experience and ensure consistency across the team, we need to create an .env.example file that lists all necessary keys without including actual sensitive values. We also need to modify the production setup so that our TAs can test deploying our application successfully.

### Acceptance Criteria

- [ ] Create a .env.example file in the root of the frontend directory.

- [ ] Identify all VITE_ variables currently used in the codebase 

- [ ] Add these keys to the file with placeholder values or descriptive comments.

- [ ] Update the README.md to include a "Getting Started" section that mentions copying .env.example to .env.

- [ ] Ensure .env remains in .gitignore.

### Meta (optional)

_No response_

sezindogan:
# 📝 Description
Fixes #645 

  The frontend has a VITE_SITE_URL environment variable used to construct absolute URLs in Schema.org JSON-LD structured data (story `@id` fields, profile url fields). Before this PR, the variable existed in the frontend source but was never passed through the Docker build pipeline, which meant it silently fell back to a hardcoded "https://storymap.page" in StructuredData/constants.js for every deployment — regardless of what domain the app was actually running on. This PR wires it properly through the build chain and adds a .env.example so new developers know it exists.

##  What VITE_SITE_URL does

  Vite bakes environment variables prefixed with VITE_ into the JavaScript bundle at build time — they are not runtime values. This means the variable must be passed as a Docker ARG during the image build, not injected later at container start.

VITE_SITE_URL is consumed in a single place:

```
// frontend/src/components/StructuredData/constants.js
export const SITE_URL = import.meta.env.VITE_SITE_URL ?? "https://storymap.page";

```
This value is embedded into the Schema.org JSON-LD <script type="application/ld+json"> blocks that are injected into story detail and profile pages. Search engines (Google, Bing) use these to understand the canonical URL of a page. If SITE_URL is wrong, the structured data points to the wrong domain.

###  How it worked before

  The variable was declared in constants.js with a ?? "https://storymap.page" fallback, but docker-compose.prod.yml and nginx/Dockerfile had no knowledge of it. Because Vite replaces import.meta.env.VITE_SITE_URL with the literal string undefined when the variable is absent at build time, the fallback always kicked in. Every production build regardless of deployment domain would embed https://storymap.page in its structured data.

### How it works now

  1. nginx/Dockerfile — declares VITE_SITE_URL as a build ARG and exports it as an ENV so Vite sees it during npm run build
  2. docker-compose.prod.yml — passes ${VITE_SITE_URL:-https://storymap.page} to the build args, reading from the host environment (i.e. .env.prod) with https://storymap.page as the default
  3. .env.prod.example — documents VITE_SITE_URL=https://yourdomain.com so operators know to set it
  4. frontend/.env.example (new file) — documents both VITE_API_URL and VITE_SITE_URL for local development, with sensible dev defaults

###  Will anything break in production?

  No. The Docker Compose default ${VITE_SITE_URL:-https://storymap.page} means that if an existing deployment doesn't set VITE_SITE_URL in its
  .env.prod, the build will behave exactly as before — the fallback is identical to the hardcoded value that was always used. No rebuild is
  required unless an operator wants to set the correct domain (which they should).

  The variable has no effect on authentication, email sending, API calls, or any runtime behaviour — it is only used in the JSON-LD structured
  data injected into story and profile pages.

  Other changes

  - README.md (root) — adds VITE_SITE_URL to the production env vars table and the "fields that must match your domain" checklist
  - frontend/README.md — comprehensive update: correct env var defaults, full route table, expanded folder structure reflecting all current
  component directories, services table listing all 16 service files, and a new E2E testing section covering all Playwright scripts
  - .env.prod.example — adds a clarifying comment on NOMINATIM_CONTACT_EMAIL explaining that Nominatim will block requests if this is not set to
   a real address

###  New features

  - frontend/.env.example — new file documenting both env vars for local dev setup

###  Modified files

  - nginx/Dockerfile — declare and export VITE_SITE_URL as a build arg so Vite bakes it into the bundle
  - docker-compose.prod.yml — pass VITE_SITE_URL to the nginx build args with https://storymap.page as fallback
  - .env.prod.example — document VITE_SITE_URL; clarify NOMINATIM_CONTACT_EMAIL requirement
  - README.md — add VITE_SITE_URL to production setup instructions
  - frontend/README.md — full documentation update: env vars, routes, folder structure, services, E2E test commands

 ### Tests

  - No code logic changed; no new unit tests required
  - Existing StructuredData tests continue to pass — they import SITE_URL directly and are unaffected by the build pipeline change

# 📊 Type of Change
- [ ]  Bug fix
- [x]  New feature
- [ ]  Refactoring
- [ ]  Documentation
 
- [ ] I have added/updated tests

# 📸 Screenshots / Recordings

# ⏱️ Estimated Review Time
- [x] < 5 min
- [ ] 5-15 min
- [ ] > 15 min

### DIFF
--- a/.env.prod.example
+++ b/.env.prod.example
@@ -9,8 +9,9 @@ DB_ROOT_PASSWORD=change-me-root
 DB_HOST=127.0.0.1
 DB_PORT=3306
 MEDIA_ROOT=/app/media
-# Passed as a Docker build arg to nginx — the React bundle will call this URL.
+# Passed as Docker build args to nginx — baked into the React bundle at build time.
 VITE_API_URL=https://yourdomain.com/api
+VITE_SITE_URL=https://yourdomain.com
 # Injected into the nginx config at container startup via envsubst.
 NGINX_HOST=yourdomain.com
 # Optional: set to your Sentry DSN to enable error tracking.
@@ -24,4 +25,5 @@ SENTRY_DSN=
 EMAIL_BACKEND=common.email_backend.ResendEmailBackend
 RESEND_API_KEY=re_your_api_key_here
 DEFAULT_FROM_EMAIL=Local History Story Map <noreply@yourdomain.com>
+# This needs to be set to a valid email address for Nominatim geocoding to work. Otherwise the requests get blocked.
 NOMINATIM_CONTACT_EMAIL=your-email@yourdomain.com
--- a/README.md
+++ b/README.md
@@ -207,6 +207,7 @@ Fill in all values. The fields that must match your domain:
 ALLOWED_HOSTS=yourdomain.com
 CORS_ALLOWED_ORIGINS=https://yourdomain.com
 VITE_API_URL=https://yourdomain.com/api
+VITE_SITE_URL=https://yourdomain.com
 NGINX_HOST=yourdomain.com
 ```
 
--- a/docker-compose.prod.yml
+++ b/docker-compose.prod.yml
@@ -40,6 +40,7 @@ services:
       dockerfile: nginx/Dockerfile
       args:
         VITE_API_URL: ${VITE_API_URL}
+        VITE_SITE_URL: ${VITE_SITE_URL:-https://storymap.page}
     restart: unless-stopped
     ports:
       - "80:80"
--- a/frontend/.env.example
+++ b/frontend/.env.example
@@ -1,0 +1,9 @@
+# Backend API base URL
+# Dev:  http://localhost:8000
+# Prod: https://your-api-domain.com
+VITE_API_URL=http://localhost:8000
+
+# Frontend public URL (used for constructing absolute links, e.g. email verification)
+# Dev:  http://localhost:5173
+# Prod: https://your-frontend-domain.com
+VITE_SITE_URL=http://localhost:5173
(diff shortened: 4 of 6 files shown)


AysuKeskin (changes requested):
Overall It looks solid, I only have one request:

`nginx/Dockerfile` declares `ARG VITE_SITE_URL` without a default. This is safe when building through `docker-compose.prod.yml`, because Compose passes `${VITE_SITE_URL:-https://storymap.page}`. However, if someone builds `nginx/Dockerfile` directly, `VITE_SITE_URL` can become an empty string, and the `?? "https://storymap.page"` fallback in `StructuredData/constants.js` will not apply because an empty string is not nullish.

Could we make the Dockerfile robust as well?

```dockerfile
ARG VITE_SITE_URL=https://storymap.page
ENV VITE_SITE_URL=$VITE_SITE_URL

sezindogan:
Valid suggestion. The default is also added to the Dockerfile, providing a more robust fallback. Ready for review.

AysuKeskin (approved):
PR now looks solid. The issue I previously pointed out has been resolved. Ready to merge.

